Issue #76, opened 2015-04-20
Verify: Save order of batched requests is guaranteed


congysu:
RESTier has an EF provider with transaction batch save support.  Verify if the following issue still exists:
Save order of batched requests is not guaranteed (e.g., could insert line-items before inserting parent order)

congysu:
Related [test case](https://github.com/OData/RESTier/blob/master/test/Microsoft.Restier.Samples.Northwind.Tests/ODataFeedTests.cs): `public async Task TestBatch()`.

congysu:
On client side, create a batch request:
1.	Create a new Customer
2.	Create a new Order: order has a foreign key to the new customer.
3.	Commit

and:
1.	Create a new Customer
2.	Create a new Order: order has a foreign key to the new customer.
3.	Rollback on server side.

The result is:
An e2e test to cover this scenario in RESTier.

lewischeng-ms:
Case added in #84

PR #84, opened 2015-05-11
Verify changeset atomicity in batch.

karataliu:


congysu:
The related issue: https://github.com/OData/RESTier/issues/76

PR #93, opened 2015-06-13
Integrate EF7 beta6

rayao:
Resolved comments in the original CR.
VS must have latest NuGet client tool to install EF7 packages.

### DIFF
--- a/src/Microsoft.Restier.EntityFramework/DbDomain.cs
+++ b/src/Microsoft.Restier.EntityFramework/DbDomain.cs
@@ -20,6 +20,7 @@ namespace Microsoft.Restier.EntityFramework
     /// <summary>
     /// Represents a domain over a DbContext.
     /// </summary>
+    [CLSCompliant(false)]
     public class DbDomain<T> : DomainBase
         where T : DbContext
     {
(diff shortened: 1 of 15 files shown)


rayao:
Forgot to mention that 6 baseline comparison cases still fail, which is expected for EF7 for now.

lewischeng-ms:
@rayao Please add ignore `.vs/` to `.gitignore` as it is generated by VS2015.

src/Microsoft.Restier.EntityFramework/DbDomain.cs
```diff
@@ -20,6 +20,7 @@ namespace Microsoft.Restier.EntityFramework
     /// <summary>
     /// Represents a domain over a DbContext.
     /// </summary>
+    [CLSCompliant(false)]
```
congysu: Why is DbDomain with `[CLSCompliant(false)]`?

src/Microsoft.Restier.EntityFramework7/Model/ModelProducer.cs
```diff
@@ -16,6 +16,7 @@
 using System.Threading;
 using System.Threading.Tasks;
 using EdmModel = Microsoft.OData.Edm.Library.EdmModel;
+using Microsoft.Data.Entity.Infrastructure;
```
congysu: Sort using

src/Microsoft.Restier.EntityFramework/DbDomain.cs
rayao: EF7 DbContext type is not CLS compliant, and that makes a type references it also not compliant. This has been modified to conditionally with EF7.

congysu: Thanks Ray.

PR #95, opened 2015-06-17
Bug fix, CLSCompliant(false) should be only applied to EF7.

rayao:
Sorry guys.


lewischeng-ms:
@rayao Looks good. Thanks.

PR #106, opened 2015-07-09
Add and enable Microsoft.OData.StyleCop custom rules

lewischeng-ms:


tools/src/Microsoft.OData.StyleCop/LineLengthRule.xml
```diff
@@ -0,0 +1,17 @@
+﻿<?xml version="1.0" encoding="utf-8" ?>
+<SourceAnalyzer Name="Microsoft.OData.StyleCop.LineLengthRule">
+  <Description>
+    Checks for lines that are too long.
+  </Description>
+  <Properties>
+    <IntegerProperty Name="LineLength" DefaultValue="-1" FriendlyName="Max Line Length"
```
karataliu: consider choose a valid default value? e.g. 80

lewischeng-ms: ok.

PR #108, opened 2015-07-13
Add LineSpaceRule to custom Style rule

lewischeng-ms:


lewischeng-ms:
Merged as https://github.com/OData/RESTier/commit/1e65a850b910bfcb53a0b9aadc1c09b36aea89ce

tools/src/Microsoft.OData.StyleCop/LineSpaceRule.cs
```diff
@@ -0,0 +1,81 @@
...
+                            csharpDocument.RootElement,
+                            lineNumber,
+                            TrailingWhiteSpacesRuleName);
+                    }
+
+                    if (checkTabIndentation)
+                    {
+                        foreach (var c in line)
+                        {
+                            if (!char.IsWhiteSpace(c))
```
congysu: Do we check `\t` in the middle?  Or it is good enough to check `\t` in the beginning.

lewischeng-ms: Here we only check leading tab indentation. If we need to check `\t` in the middle, we should set  up another rule to do this.

PR #109, opened 2015-07-14
Fix StyleCop issues that belong to CSharp.Layout

lewischeng-ms:


src/Microsoft.Restier.Core/Submit/ChangeSet.cs
```diff
@@ -52,7 +53,7 @@ public IList<ChangeSetEntry> Entries
         /// <summary>
         /// Gets or sets a value indicating whether an Entity has been added, modified, or deleted.
         /// </summary>
-        // TODO GitHubIssue#37 : make the ChangeSet 'dynamic' so it gets added to as things change during the flow
+        /// TODO GitHubIssue#37 : make the ChangeSet 'dynamic' so it gets added to as things change during the flow
```
congysu: `// TODO` on top of the `/// <summary>`?  Therefore, XML comments translation won't translate it.

lewischeng-ms: Good idea.

PR #110, opened 2015-07-14
Fix invalid spacing

lewischeng-ms:


src/Common.Stylecop
```diff
@@ -61,6 +61,15 @@
         </Rule>
       </Rules>
     </Analyzer>
+    <Analyzer AnalyzerId="Microsoft.StyleCop.CSharp.MaintainabilityRules">
+      <Rules>
+        <Rule Name="FileMayOnlyContainASingleClass">
```
congysu: Is this to suppress SA1402?

lewischeng-ms: Yes, there are many cases in RESTier to do so. Shall we follow this to separate those files to contain only one class per file?

congysu: Ok to me to suppress it for now.  Thanks.

src/Microsoft.Restier.WebApi/ODataDomainController.cs
```diff
@@ -50,7 +50,7 @@ protected ODataDomainController()
         }
 
         /// <summary>
-        /// Gets the <see cref="IDomain"/> interface of the domain of the <see cref="ODataDomainController"/>.
```
congysu: What is the issue with `<see cref="*`?

lewischeng-ms: @karataliu suggested that "Gets the interface of ..." is not proper so change to this statement.
P.S., this comment was originally not present and added by me several days ago. So it's OK to change the way expressing.

PR #111, opened 2015-07-14
Fix StyleCop issues that belong to CSharp.Readability

lewischeng-ms:


src/Shared/TypeExtensions.cs
```diff
@@ -32,5 +40,10 @@ public static Type FindGenericType(this Type type, Type definition)
 
             return null;
         }
+
+        internal static MethodInfo GetQualifiedMethod(this Type type, string methodName)
```
congysu: `public`? given that the type `TypeExtensions` is `internal`.

src/Shared/TypeExtensions.cs
```diff
@@ -1,10 +1,18 @@
...
 // Licensed under the MIT License.  See License.txt in the project root for license information.
 
+using System.Reflection;
+
 namespace System
 {
     internal static partial class TypeExtensions
     {
+        private const BindingFlags QualifiedMethodBindingFlags = BindingFlags.NonPublic |
+                                                                 BindingFlags.Static |
```
congysu: Is this indentation suggested by StyleCop.  Seems different from method parameters indentation.

lewischeng-ms: Actually, StyleCop has no preference on the indentation of initialization expressions. Just use this style to make it look prettier than the indentation as parameters.

src/Shared/TypeExtensions.cs
lewischeng-ms: Nice catch. I will update this in some consecutive commit.

PR #144, opened 2015-07-22
Update version to 0.3.0-beta

karataliu:


License.txt
```diff
@@ -1,4 +1,4 @@
-RESTier ver. 0.2-pre
+RESTier ver. 0.3-beta
```
congysu: Can remove ` ver. 0.*` in this line.

Issue #116, opened 2015-07-17
[Protocol] StoreGeneratedPattern should not appear for Identity


karataliu:
ModuleProducer.cs:

```
   MetadataProperty storeGeneratedPattern = null;
                    efProperty.MetadataProperties.TryGetValue(
                        AnnotationSchema + ":StoreGeneratedPattern",
                        true,
                        out storeGeneratedPattern);

                    if (storeGeneratedPattern != null)
                    {
                        SetComputedAnnotation(model, property);
                    }
```

This would generate StoreGeneratedPattern even the value is Identity.

Should verifiy the value is Computed when setting such pattern.

karataliu:
```
<Term Type="Core.Tag" Name="Computed" AppliesTo="Property" DefaultValue="true">
<Annotation Term="Core.Description" String="A value for this property is generated on both insert and update"/>
</Term>
```

https://msdn.microsoft.com/en-us/library/system.data.metadata.edm.storegeneratedpattern(v=vs.110).aspx

lewischeng-ms:
Resolved in #152

PR #152, opened 2015-07-31
Various fixes for RESTier GA

lewischeng-ms:


lewischeng-ms:
@karataliu Please help review. Thanks.

karataliu:
looks good

src/Microsoft.Restier.EntityFramework/Model/ModelProducer.cs
```diff
@@ -193,7 +193,7 @@ private ModelProducer()
                         true,
                         out storeGeneratedPattern);
 
-                    if (storeGeneratedPattern != null)
+                    if (storeGeneratedPattern != null && (string)storeGeneratedPattern.Value == "Computed")
```
congysu: Is this for #116?  @lewischeng-ms @challenh Can we have a sync on this?  How to represent `StoreGeneratedPattern.Identity` with OData annotation?

lewischeng-ms: Actually according to OData [standard vocabularies](http://docs.oasis-open.org/odata/odata/v4.0/errata02/os/complete/vocabularies/Org.OData.Core.V1.xml), we don't support any annotation regarding StoreGeneratedPattern.Identity but only Computed.

PR #159, opened 2015-07-31
Supporting any IDbSet<> implementations

mkemal:
Currently, restier does not support IDbSet<> implementations other than DbSet<>. So, if one implements a custom IDbSet<> implementation, or even extends the default implementation, DbSet<>, he/she would not be able to use it with RESTier. Because, as far as i know by inspecting the ModelMapper implementation, any IDbSet<> implementation not being DbSet<> is rejected.

For example, when using IdentityDbContext, because it defines the entity sets as IDbSet<>, i have not be able to work it with Restier.

To overcome this problem, i created my custom IModelMapper by just copying the default one, and adding the changes i included in this commit. Currently i am using this implementation and I don't know if there is any reason for this constraint in the default implementation. So, instead of asking you if this is a proper solution, and explaining what i did in details, i decided to make a pull request.

If it may lead any problems with the rest of the framework, please let me know about it. Because, as i said, currently i use this implementation with my custom IModelMapper implementation which i created just to overcome this problem.

### DIFF
--- a/src/Microsoft.Restier.EntityFramework/Model/ModelMapper.cs
+++ b/src/Microsoft.Restier.EntityFramework/Model/ModelMapper.cs
@@ -50,7 +50,7 @@ namespace Microsoft.Restier.EntityFramework.Model
         /// <c>true</c> if the relevant type was
         /// provided; otherwise, <c>false</c>.
         /// </returns>
-        public bool TryGetRelevantType(
+        public virtual bool TryGetRelevantType(
             DomainContext context,
             string name,
             out Type relevantType)
@@ -61,42 +61,14 @@ namespace Microsoft.Restier.EntityFramework.Model
             if (property != null)
             {
                 var type = property.PropertyType;
-                var parent = GetGenericParent(type);
-                if (parent != null)
+                var genericType = type.FindGenericType(typeof (IDbSet<>));
+                if (genericType != null)
                 {
-                    relevantType = parent.GetGenericArguments()[0];
+                    relevantType = genericType.GetGenericArguments()[0];
                 }
             }
 
             return relevantType != null;
-        }
-
-        /// <summary>
-        /// Finds the <c>IDbSet</c> interface this type implements.
-        /// </summary>
-        /// <param name="type">
-        /// The type of the entity set.
-        /// </param>
-        /// <returns>
-        /// the type itself if it is defined as DbSet or IDbSet,
-        /// type of IDbSet interface implemented by this type if there is;
-        /// otherwise, null
-        /// </returns>
-        private static Type GetGenericParent(Type type)
-        {
-            // Because usage of DbSet very common, first check for it to speed things up
-            if (type.IsGenericType)
-            {
-                var generic = type.GetGenericTypeDefinition();
-                if (generic == typeof(DbSet<>) || generic == typeof(IDbSet<>))
-                {
-                    return type;
-                }
-            }
-
-            return
-                type.GetInterfaces()
-                    .FirstOrDefault(t => t.IsGenericType && t.GetGenericTypeDefinition() == typeof (IDbSet<>));
         }
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


karataliu:
@mkemal 
Thanks for the contribution! The change looks good in general.

Can you help fix the following issues:
1. The code won't compile, unless System.Linq namespace using added;
2. There are some build warnings from FxCop, please clean up all warnings.

mkemal:
@karataliu 

I had copied the changes from my custom implementation. Because i didn't even open the project in VS I have never compiled it before submiting. So, i had forgetten to include Linq namespace import. Sorry for that.

1. I added the using statement for the Linq namespace.
2. I run the VS 2013 Code Analyze and converted the 'GetGenericParent' method to static, according to analysis results.

src/Microsoft.Restier.EntityFramework/Model/ModelMapper.cs
```diff
@@ -60,10 +61,10 @@ public ModelMapper(Type dbContextType)
             if (property != null)
             {
                 var type = property.PropertyType;
-                if (type.IsGenericType &&
-                    type.GetGenericTypeDefinition() == typeof(DbSet<>))
+                var parent = GetGenericParent(type);
```
congysu: There is a `TypeExtensions.FindGenericType` method: https://github.com/OData/RESTier/blob/993fdc85cb62919d7f714f65bfe344884cfeee2f/src/Shared/TypeExtensions.cs#L16

mkemal: It can be used, however implementation of the `TypeExtensions.FindGenericType` seems a bit costly in terms of computational complexity. Say the type represented by the parameter `type` is at level `h` in its inheritance tree (distance from the type of `object`) and there are `n` interfaces implemented by it. Here are some cases:

- If the parameter `definition` does not represent an interface, then the complexity of the method would be `O(h)`,
- If the parameter `definition` represents an interface implemented by `type`, then the complexity of the method would be `O(n)`
- However, if the parameter `definition` represents an interface which is not implemented by `type`, then the complexity would be `O(n h)` which represents the worst case complexity of this method.

To make the worst case complexity `O(max(n, h))`, at least the inner `foreach` loop which checks the `definition` against the implemented interfaces should be moved out of the while loop.

If it is OK for you, i can add the changes proposed above and for its usage in `ModelMapper` to this pull request.

congysu: It is OK for me - C# is with single inheritance, and `GetInterfaces` returns all interfaces the class/interface implements, and inherits.  Please proceed with the change.  Thanks.

mkemal: I pushed the changes on `FindGenericType` method. However, `TypeExtensions` class appears to be marked internal, as it is defined in `System` namesapce. So, i couldn't use it from the `ModelMapper` class. It seems to me, the best solution is moving `TypeExtensions` class out from `System` namespace and make it public. However, as it will effect all places where the method is used, i didn't take any action.

congysu: You can link the `TypeExtensions` class to  `Microsoft.Restier.EntityFramework.csproj`.  An example: https://github.com/OData/RESTier/blob/master/src/Microsoft.Restier.EntityFramework/Microsoft.Restier.EntityFramework.csproj#L40-L45

mkemal: Thanks for the trick; it is good to know the usage of such compile time dependencies. I pushed the changes.

src/Shared/TypeExtensions.cs
```diff
@@ -13,34 +13,82 @@ internal static partial class TypeExtensions
                                                                  BindingFlags.IgnoreCase |
                                                                  BindingFlags.DeclaredOnly;
 
+
```
congysu: Thanks for adding comments.  Code comments are optional for internal/private class and/or method.

karataliu:
merged this change. Thanks @mkemal .

PR #198, opened 2015-08-25
Refactor of MapODataDomainRoute and ODataDomainController

lewischeng-ms:


src/Microsoft.Restier.WebApi/HttpRequestMessageExtensions.cs
```diff
@@ -40,5 +42,42 @@ public static ODataDomainChangeSetProperty GetChangeSet(this HttpRequestMessage
...
+                domainFactory = value as Func<IDomain>;
+                Contract.Assert(domainFactory != null);
+            }
+            else
+            {
+                domainFactory = null;
+            }
+
+            return domainFactory;
+        }
```
karataliu: consider simplify this method, TryGet then cast.

lewischeng-ms: Fixed in https://github.com/OData/RESTier/commit/2b6607439d3b5bcc0956b7ed492caa5ace04cdb5

Issue #234, opened 2015-09-22
[Design] Support property access in ODataDomainController


lewischeng-ms:
like ~/People(0)/UserName

lewischeng-ms:
We still need to support property routing of collection-valued properties, enum properties.

PR #236, opened 2015-09-22
Support property access in ODataDomainController

lewischeng-ms:
Not supported features: Enum and Collection<NonEntity>


lewischeng-ms:
Related issue #234

Issue #245, opened 2015-10-22
Fix EF7 project related issues


karataliu:
1. Model builder:
src/Microsoft.Restier.EntityFramework7/Model/ModelProducer.cs::CreteEntityType
2. StyleCop warnings
3. Test cases failure.

karataliu:
#1 done with https://github.com/OData/RESTier/commit/55d573454cc109fa57b18cf45c962f19ac4c3135

chinadragon0515:
Close the issue as Ray has helped fixed EF7 projects related issue.

PR #246, opened 2015-10-22
Fix EF7 project related issues

karataliu:
Fix EF7 related projects, some tests and warning still exist. Created https://github.com/OData/RESTier/issues/245 for it.


karataliu:
opening on wrong branch.

PR #254, opened 2015-11-06
EF7 ChangeSetPreparer.PrepareAsync is using the wrong ApiContext Property

eerhardt:
EF7 ChangeSetPreparer.PrepareAsync is using the wrong ApiContext.GetProperty name to find the current DbContext. It is currently hard-coded to "DbContext", when it should be using the DbApiConstants.DbContextKey constant.

This causes a NullReferenceException during change set submission because it can't find the dbContext.


lewischeng-ms:
@eerhardt Merged. Thanks for fixing the issue.

Issue #273, opened 2016-01-11
Map System.DateTime to Edm.DateTimeOffset


congysu:
Map System.DateTime to Edm.DateTimeOffset.  Leverage OData v4 Web API to map System.DateTime to Edm.Date, and Edm.TimeOfDay etc.
Related issue: https://github.com/OData/WebApi/issues/118

lewischeng-ms:
Related PR: #281.
Bug fixes: https://github.com/OData/RESTier/commit/968e82d6783e3cbb8190674f1dbc49bf3c63bc55

PR #278, opened 2016-01-19
Add DateTime/Date/TimeOfDay support. Simplify EF7 entity update logic.

rayao:
https://github.com/OData/RESTier/issues/273

Also, removed EF.SqlServer dependency from Restier.EF7


src/Microsoft.Restier.EntityFramework/Model/ModelProducer.cs
```diff
@@ -336,11 +336,46 @@ private static IEdmPrimitiveTypeReference GetPrimitiveTypeReference(EdmProperty
...
+                        {
+                            object config = metadata.Value;
+                            var columnType = (string)config.GetType().GetProperty("ColumnType").GetValue(config);
+
+                            if (string.Equals(columnType, "date", StringComparison.OrdinalIgnoreCase))
+                            {
+                                return EdmCoreModel.Instance.GetDate(efProperty.Nullable);
+                            }
+                        }
+                    }
```
lewischeng-ms: Is your code written specially for EF7 here? I am not sure about the difference that in my code the DB type name comes from the `ColumnAttribute`. But I guess we can keep both code here and use a compilation constant to differentiate EF6 and EF7?

src/Microsoft.Restier.EntityFramework/Submit/ChangeSetPreparer.cs
```diff
@@ -209,6 +209,24 @@ private static object ConvertIfNecessary(Type type, object value)
                 return (DateTime)dateValue;
             }
 
+            // Convert to System.DateTime supported by EF from DateTimeOffset.
+            if (value is DateTimeOffset)
```
lewischeng-ms: We have almost identical code here:) I just need to move some of my comments here later after your commit.

src/Microsoft.Restier.EntityFramework7/Model/ModelProducer.cs
```diff
@@ -202,6 +203,29 @@ private static void SetComputedAnnotation(EdmModel model, IEdmProperty target)
             {
                 return null;
             }
+
+            if (Type.GetTypeCode(propertyType) == TypeCode.DateTime)
```
lewischeng-ms: I remember some .NET profiles have retired `Type.GetTypeCode` thus we copied the whole `TypeCode` enum into ODataLib. Is there any reason we are not using `propertyType == typeof(DateTime)` here?

src/Shared/TypeExtensions.cs
```diff
@@ -105,4 +103,62 @@ private static bool IsGenericDefinition(this Type type, Type definition)
                    type.GetGenericTypeDefinition() == definition;
         }
     }
+
+    internal static class TypeHelper
+    {
+        public static Type GetUnderlyingTypeOrSelf(Type type)
+        {
+            return Nullable.GetUnderlyingType(type) ?? type;
```
lewischeng-ms: Great! I forgot to take nullable temporal types into account when comparing type in `ChangeSetPreparer` and the converters. I will change to use this helper class for other type comparisons later.

src/Microsoft.Restier.Samples.Northwind/Models/NorthwindContext.cs
```diff
@@ -121,7 +121,7 @@ protected override void OnModelCreating(Data.Entity.ModelBuilder modelBuilder)
             modelBuilder.Entity<Order>(entityBuilder =>
             {
                 entityBuilder.ToTable("Orders");
-                entityBuilder.HasMany(e => e.Order_Details).WithOne(e => e.Order).Required().ForeignKey(e => e.OrderID);
+				entityBuilder.HasMany(e => e.Order_Details).WithOne(e => e.Order).Required().ForeignKey(e => e.OrderID);
```
lewischeng-ms: please use space instead of tab in source code.

src/Microsoft.Restier.EntityFramework7/Submit/ChangeSetPreparer.cs
```diff
@@ -116,78 +126,110 @@ private ChangeSetPreparer()
...
+        private static void SetValues(EntityEntry dbEntry, DataModificationEntry entry, Type entityType)
+        {
+            foreach (KeyValuePair<string, object> propertyPair in entry.LocalValues)
             {
-                // The algorithm for a "FullReplaceUpdate" is taken from WCF DS ObjectContextServiceProvider.ResetResource, and is as follows:
-                // Create a new, blank instance of the entity.  Copy over the key values, and set any updated values from the client on the new instance.
-                // Then apply all the properties of the new instance to the instance to be updated.  This will set any unspecified
-                // properties to their default value.
+                PropertyEntry propertyEntry = dbEntry.Property(propertyPair.Key);
+                Type type = propertyEntry.CurrentValue.GetType();
```
lewischeng-ms: Actually here is a bug because the current value could be null thus there is no type here. I've fixed the bug in EF6 project but not EF7. I think we shall have a task #280 to sync the latest features from EF6 to EF7.

src/Microsoft.Restier.EntityFramework7/Model/ModelProducer.cs
```diff
@@ -202,6 +203,29 @@ private static void SetComputedAnnotation(EdmModel model, IEdmProperty target)
...
+                if (string.Equals(columnType, "date", StringComparison.OrdinalIgnoreCase))
+                {
+                    kind = EdmPrimitiveTypeKind.Date;
+                }
+            }
+
+            if (propertyType == typeof(TimeSpan))
+            {
+                RelationalPropertyAnnotations annotations = new RelationalPropertyAnnotations(efProperty, null);
+                var columnType = annotations.ColumnType;
```
lewischeng-ms: Is this new feature in EF7's `efProperty`?

src/Microsoft.Restier.EntityFramework/Model/ModelProducer.cs
rayao: No this works only for EF6, EF7 has a much cleaner approach.
EF7 shares most code files from EF6, but has dedicated ModelProducer and ChangeSetPreparer.

src/Microsoft.Restier.EntityFramework7/Model/ModelProducer.cs
rayao: Oh it's my fault, this is written before I brought TypeHelper in, will change it.

src/Microsoft.Restier.EntityFramework7/Model/ModelProducer.cs
rayao: Yeah this relies on EF7 annotations, EF7 model configurations work in this way.

src/Microsoft.Restier.EntityFramework/Submit/ChangeSetPreparer.cs
rayao: If you like I can revert all EF6 changes to make you merge easier.

rayao:
Hi Lewis, perhaps I'd better remove all EF6 changes to make your merge easier? Or you can check in first and let me merge.

lewischeng-ms:
@rayao, my PR need to be based on yours and use some code you wrote. So please address the comments and merge this PR first if all tests can pass and there is no code analysis issue during compilation. Thanks!

rayao:
Sorry I messed up the branch hierarchy. Didn't notice there're check-ins before merge. Since it tells me "no conflict", I took that for "no change since last sync".
I'm not very familiar with git commands, I'm afraid of messing it up further, could someone help restore a single lineage?

lewischeng-ms:
@rayao Don't worry about that. You didn't mess up the master branch history. It's OK to merge a pull request like this. For future, if you want to make the commits a single line, you can first do a `git pull --rebase OData master` to pull and rebase all the latest changes from `OData/master` then do a `git push OData master` to check in.

PR #283, opened 2016-01-22
EF7: Pass code analysis check. Cascade deletion support.

rayao:
* Format text to meet FxCop requirements.
* Add cascade deletion configuration since EF7 has support now.

Verified that UTs passed for EF6.


lewischeng-ms:
Thanks for the PR. Looks good to me.

PR #285, opened 2016-01-31
Microsoft.FX.DI integration [Option 1]

rayao:
Use Micorsoft DI framework to replace current Dictionary<,> based service registration.
This is a relatively aggressive change, compared to my previous PR. There's an option 2 #286 pull request which is basically identical to the previous PR #264. It's up to you to make a choice.
In this change the main difference is that class **ApiBuilder** is added for API service registration, and later build an ApiConfiguration.
Pros:
- No **Committed** state for ApiConfiguration, it's committed once created. For now those related methods are not removed yet but they do nothing.
- ApiConfiguration itself is registered as a service and participate in DI. That means a descendant Api class could register some service which gets an ApiConfiguration injected on construction.

But of course, at the same time this change inevitably has broader impact to existing code.
- ApiConfiguration public interface changes, This breaks a lot of things, which of course are fixed with this change. But, any consumer out there may have a bigger chance to break, compared to option 2.

### DIFF
--- a/src/Microsoft.Restier.Core/ApiBase.cs
+++ b/src/Microsoft.Restier.Core/ApiBase.cs
@@ -64,10 +64,13 @@ namespace Microsoft.Restier.Core
                     ApiConfiguration configuration;
                     if (!Configurations.TryGetValue(apiType, out configuration))
                     {
-                        configuration = this.CreateApiConfiguration();
-                        EnableConventions(configuration, apiType);
-                        ApiConfiguratorAttribute.ApplyConfiguration(
-                            apiType, configuration);
+                        var builder = this.ConfigureApiBuilder(new ApiBuilder());
+                        EnableConventions(builder, apiType);
+                        ApiConfiguratorAttribute.ApplyApiBuilder(apiType, builder);
+                        builder.TryUseSharedApiScope(); // TODO: Maybe default to context scope?
+
+                        configuration = this.CreateApiConfiguration(builder);
+                        ApiConfiguratorAttribute.ApplyConfiguration(apiType, configuration);
                         Configurations[apiType] = configuration;
                     }
 
@@ -121,14 +124,31 @@ namespace Microsoft.Restier.Core
         }
 
         /// <summary>
-        /// Creates the API configuration for this API.
+        /// Configure services for this API.
         /// </summary>
-        /// <returns>
-        /// The API configuration for this API.
-        /// </returns>
-        protected virtual ApiConfiguration CreateApiConfiguration()
+        /// <param name="builder">
+        /// The <see cref="ApiBuilder"/> with which to create an <see cref="ApiConfiguration"/>.
+        /// </param>
+        /// <returns>The <see cref="ApiBuilder"/>.</returns>
+        [CLSCompliant(false)]
+        protected virtual ApiBuilder ConfigureApiBuilder(ApiBuilder builder)
         {
-            return new ApiConfiguration();
+            return builder;
+        }
+
+        /// <summary>
+        /// Creates the API configuration for this API.
+        /// Descendants may override to use a customized DI container, or further configure the built
+        /// <see cref="ApiConfiguration"/>.
+        /// </summary>
+        /// <param name="builder">The <see cref="ApiBuilder"/> containing API service registrations.</param>
+        /// <returns>
+        /// An <see cref="ApiConfiguration"/> with which to create the API configuration for this API.
+        /// </returns>
+        [CLSCompliant(false)]
+        protected virtual ApiConfiguration CreateApiConfiguration(ApiBuilder builder)
+        {
+            return builder.Build();
         }
 
         /// <summary>
@@ -156,6 +176,11 @@ namespace Microsoft.Restier.Core
         /// </param>
         protected virtual void Dispose(bool disposing)
         {
+            if (this.apiContext != null)
+            {
+                this.apiContext.DisposeScope();
+            }
+
             if (disposing)
             {
                 this.apiContext = null;
@@ -166,7 +191,7 @@ namespace Microsoft.Restier.Core
         /// <summary>
         /// Enables code-based conventions for an API.
         /// </summary>
-        /// <param name="configuration">
+        /// <param name="builder">
         /// An API configuration.
         /// </param>
         /// <param name="targetType">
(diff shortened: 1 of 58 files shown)


lewischeng-ms:
@rayao Thanks so much for the PR. Personally I also prefer your **option 1**: it makes much sense to treat `ApiConfiguration` as a "service" and manage all services consistently in DI. Since now `ApiConfiguration` itself participates in DI, `ApiBuilder` plays an important role on decoupling  `ApiConfiguration` from other services (hook handlers). For this reason, I would suggest we do this more aggressively and completely and make the API cleaner. I will comment more on some specific lines.

rayao:
@lewischeng-ms Thank you Lewis, really happy that we have common thoughts. Since Spring Festival is approaching, if you're going home before the official holiday, don't urge yourself to finish this PR before you leaving, just go at your own pace, it's OK for me to wait. I just want to be helpful and really hate to become a burden of any of you. :)

src/Microsoft.Restier.Core/ApiBase.cs
```diff
@@ -121,14 +124,31 @@ public void Dispose()
...
 
         /// <summary>
+        /// Configure services for this API.
+        /// </summary>
+        /// <param name="builder">
+        /// The <see cref="ApiBuilder"/> with which to create an <see cref="ApiConfiguration"/>.
+        /// </param>
+        /// <returns>The <see cref="ApiBuilder"/>.</returns>
+        [CLSCompliant(false)]
+        protected virtual ApiBuilder ConfigureApiBuilder(ApiBuilder builder)
```
lewischeng-ms: Suggest a more concise name `ConfigureApi` because we are actually configuring and finally building an `Api` here.

rayao: Sure, thanks. also If any other naming is inappropriate please let me know.

src/Microsoft.Restier.Core/ApiBuilderExtensions.cs
```diff
@@ -0,0 +1,564 @@
...
+                if (!e.MoveNext())
+                {
+                    return null;
+                }
+
+                T first = e.Current.Instance;
+                T current = first;
+                while (e.MoveNext())
+                {
+                    var delegateHandler = current as IDelegateHookHandler<T>;
```
lewischeng-ms: #289

src/Microsoft.Restier.Core/ApiBuilderExtensions.cs
```diff
@@ -0,0 +1,564 @@
...
+            public IServiceScope CreateApiScope()
+            {
+                return this.Factory.CreateScope();
+            }
+        }
+    }
+
+    internal static class ChainedService
+    {
+        public static Tuple<object, ConstructorInfo> CreateViaReflection(
```
lewischeng-ms: Design discussion: This part of logic duplicates with the functionality of DI. How about the following policy?

 - ctor. We don't care about how the ctor of a hook handler is implemented. We just let DI manage the instantiation itself.
 - public property `InnerHandler`. We ask the user to write such a public property in his hook handler class if the hook handler is supposed to be chained. Ignore the property if there is no such need. Then we use the similar logic in `BuildLegacyHandlers` to create the hook handler instances.

src/Microsoft.Restier.EntityFramework/DbApi.cs
```diff
@@ -48,19 +48,23 @@ protected T DbContext
...
-            var configuration = base.CreateApiConfiguration();
-            configuration.AddHookHandler<IModelBuilder>(ModelProducer.Instance);
-            configuration.AddHookHandler<IModelMapper>(new ModelMapper(typeof(T)));
-            configuration.AddHookHandler<IQueryExpressionSourcer>(QueryExpressionSourcer.Instance);
-            configuration.AddHookHandler<IQueryExecutor>(QueryExecutor.Instance);
-            configuration.AddHookHandler<IChangeSetPreparer>(ChangeSetPreparer.Instance);
-            configuration.AddHookHandler<ISubmitExecutor>(SubmitExecutor.Instance);
-            return configuration;
+            builder = base.ConfigureApiBuilder(builder);
+            builder.AddHookHandler<IModelBuilder>(ModelProducer.Instance);
```
lewischeng-ms: #290

src/Microsoft.Restier.Core/ApiBuilderExtensions.cs
rayao: UP! @lewischeng-ms

src/Microsoft.Restier.Core/ApiBuilder.cs
```diff
@@ -0,0 +1,60 @@
+﻿// Copyright (c) Microsoft Corporation.  All rights reserved.
+// Licensed under the MIT License.  See License.txt in the project root for license information.
+
+using System;
+using System.Linq;
+using Microsoft.Framework.DependencyInjection;
```
lewischeng-ms: @rayao I am not sure if there is any BIG gap between beta8, rc1 and even the latest version. If so, do we consider use rc1 or latest packages now to avoid any big change in the future?

src/Microsoft.Restier.Core/ApiBuilder.cs
```diff
@@ -0,0 +1,60 @@
...
+    /// <param name="next">
+    /// Return the result of the previous contributor on the chain.
+    /// </param>
+    /// <returns>A service instance of <typeparamref name="T"/>.</returns>
+    public delegate T ApiServiceContributor<T>(IServiceProvider serviceProvider, Func<T> next) where T : class;
+
+    /// <summary>
+    /// Builder object to create an <see cref="ApiConfiguration"/>
+    /// </summary>
+    [CLSCompliant(false)]
```
lewischeng-ms: I see many occurrences of this attributes in the code. Is it a must for our library? They look somewhat difficult to maintain... Do we have any guideline that which language feature is NOT common for all .NET languages?

src/Microsoft.Restier.Core/ApiConfiguratorAttribute.cs
```diff
@@ -115,6 +145,22 @@ protected ApiConfiguratorAttribute()
...
+        /// Configures an API builder.
+        /// </summary>
+        /// <param name="builder">
+        /// An API builder.
+        /// </param>
+        /// <param name="type">
+        /// The API type on which this attribute was placed.
+        /// </param>
+        [CLSCompliant(false)]
+        public virtual void ConfigureBuilder(
```
lewischeng-ms: We can also name it `ConfigureApi` and merge with the `Configure` method below.

src/Microsoft.Restier.Core/ApiConfiguration.cs
```diff
@@ -35,79 +33,56 @@ namespace Microsoft.Restier.Core
...
+
         /// <summary>
-        /// Adds a hook handler instance.
+        /// Ensures this API configuration has been committed.
         /// </summary>
-        /// <typeparam name="T">The hook handler interface.</typeparam>
-        /// <param name="handler">An instance of hook handler for TContext.</param>
-        /// <returns>Current <see cref="ApiConfiguration"/></returns>
-        public ApiConfiguration AddHookHandler<T>(T handler) where T : class, IHookHandler
+        public void EnsureCommitted()
```
lewischeng-ms: #291

lewischeng-ms:
@rayao Apart from issue #289, #290, #291 and the design proposal about the inner handler, would you please change your PR accordingly and merge first? For the issues mentioned and subsequent improvement, we can do this gradually by sending new PRs and reviewing together. This BIG PR really makes me feel hard to review and merge other fixes... Thanks:-)

src/Microsoft.Restier.Core/ApiBuilder.cs
rayao: Yeah I know it's really troublesome. My guess is that MS.FX.DI assembly is not marked as CLSCompliant, so every type from it is not compliant. Any suggestion to circumvent?

src/Microsoft.Restier.Core/ApiConfiguratorAttribute.cs
rayao: Sure. But I'm afraid merge with the Configure is not possible, since ApiConfiguration is built from ApiBuilder, they can't be configurable at the same time.

src/Microsoft.Restier.Core/ApiBuilder.cs
rayao: The only reason to stay on beta8 is EntityFramework.SqlServer, it's not in RC yet, and it's very weird that EF7.SqlServer beta8 can't work with RC1 build of other EF modules. And since our Northwind (on EF7) sample relies on it. I suggest we do this after EF7.SqlServer releases.

rayao:
Ok, thanks!

src/Microsoft.Restier.Core/ApiBuilder.cs
rayao: And the other thing is EF team will change the package name in near future, so anyway we have to make a change to adapt later. So let's skip it this time.

src/Microsoft.Restier.Core/ApiBuilder.cs
rayao: Make ApiBuilder CLSCompliant, so that all other occurences are no longer necessary. Thank you!

rayao:
Pushed with c7338fe. Rolling build passed. :)

PR #286, opened 2016-01-31
Microsoft.FX.DI integration [Option 2]

rayao:
Use Micorsoft DI framework to replace current Dictionary<,> based service registration.
This is basically identical to my previous  #264. Compared to option 1 #285, this is a relatively conservative change. It's up to you to make a choice.
In this change ApiConfiguration has its service registration/resolve built upon MS.FX.DI, almost all public interfaces remain same.
There's 1 major breaking change:
Services or hook handlers can't be resolved before ApiConfiguraion committed.

Personally I like option 1 more, IMO it's cleaner, more componentized, and the building pattern mimics what's in ASP.NET MVC.


rayao:
Go with option 1.

PR #294, opened 2016-02-18
Issue 123 - Int64 Primary Keys

SaucyJack:
Altered the expression construction in RestierQueryBuilder to use
Convert.ChangeType for key comparisons.


lewischeng-ms:
Merged as https://github.com/OData/RESTier/commit/2923f797304eb3fb823a54ba74f2c5de9f2e07f0. Thanks for the contribution.

PR #295, opened 2016-02-19
Fix TFS build break

lewischeng-ms:
1. The digital signature of DI 1.0.0-beta8 is expired since 2/1/2016 which will fail strong name validation.
2. The TFS build machine is still using VS2013 so we have to remove the C# 6.0 language features in order to compile.

### DIFF
--- a/src/Microsoft.Restier.Core/ApiBuilder.cs
+++ b/src/Microsoft.Restier.Core/ApiBuilder.cs
@@ -3,7 +3,7 @@
 
 using System;
 using System.Linq;
-using Microsoft.Framework.DependencyInjection;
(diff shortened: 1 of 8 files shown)


lewischeng-ms:
@rayao These are urgent checkins to fix TFS build. Sorry to cause you troubles. Please rebase your PR on the latest master branch. Thanks.

rayao:
Sorry I didn't realize these problems.
Please don't hesitate to check in and let me do the merge, I'm much more familiar with git commands now. :)

lewischeng-ms:
@rayao No worries. They are easy to fix. We also expect build machine to upgrade to VS2015 so that we can take advantage of some newest C# 6.0 features.

PR #296, opened 2016-02-19
Inner service instance injection.

rayao:
Inner service instance injection. via public settable property.
 Remove EnsureCommitted & IsCommitted.
Adds a remark section for the newly added ChianPrevious method.
Merge master branch.

### DIFF
--- a/src/Microsoft.Restier.Core/ApiBuilderExtensions.cs
+++ b/src/Microsoft.Restier.Core/ApiBuilderExtensions.cs
@@ -2,7 +2,6 @@
 // Licensed under the MIT License.  See License.txt in the project root for license information.
 
 using System;
-using System.Collections.Generic;
 using System.Linq;
 using System.Linq.Expressions;
 using System.Reflection;
@@ -228,8 +227,9 @@ namespace Microsoft.Restier.Core
                     return instance;
                 }
 
-                var nextProperty = typeof(TImplement).GetTypeInfo().GetProperties()
-                    .FirstOrDefault(e => e.SetMethod.IsPublic && e.PropertyType == typeof(TService));
+                var nextProperty = typeof(TImplement).GetTypeInfo()
+                    .GetProperties(BindingFlags.Instance | BindingFlags.Public | BindingFlags.NonPublic)
+                    .FirstOrDefault(e => e.SetMethod != null && e.PropertyType == typeof(TService));
                 if (nextProperty == null)
                 {
                     factory = (serviceProvider, _) => serviceProvider.GetRequiredService<TImplement>();
@@ -239,6 +239,13 @@ namespace Microsoft.Restier.Core
                 nextProperty.SetValue(instance, next());
                 factory = (serviceProvider, getNext) =>
                 {
+                    // To build a lambda expression like:
+                    // (sp, next) => 
+                    // {
+                    //     var hook = sp.GetRequiredService<TImplement>();
+                    //     hook.next = next();
+                    //     return hook;
+                    // }
                     var serviceProviderParam = Expression.Parameter(typeof(IServiceProvider));
                     var nextParam = Expression.Parameter(typeof(Func<TService>));
 
@@ -431,8 +438,8 @@ namespace Microsoft.Restier.Core
             {
                 return this.Factory.CreateScope();
             }
-                    }
-                    }
+        }
+    }
 
     internal static class ChainedService<T> where T : class
     {
--- a/src/Microsoft.Restier.Core/ApiBuilderExtensions.cs
+++ b/src/Microsoft.Restier.Core/ApiBuilderExtensions.cs
@@ -240,7 +240,7 @@ namespace Microsoft.Restier.Core
                 factory = (serviceProvider, getNext) =>
                 {
                     // To build a lambda expression like:
-                    // (sp, next) => 
+                    // (sp, next) =>
                     // {
                     //     var hook = sp.GetRequiredService<TImplement>();
                     //     hook.next = next();
--- a/src/Microsoft.Restier.Core/ApiBuilderExtensions.cs
+++ b/src/Microsoft.Restier.Core/ApiBuilderExtensions.cs
@@ -152,11 +152,7 @@ namespace Microsoft.Restier.Core
             Ensure.NotNull(obj, "obj");
             Ensure.NotNull(contributor, "contributor");
 
-            // Services have singleton lifetime by default, call Make... to change.
-            obj.Services.TryAddSingleton(typeof(T), ChainedService<T>.DefaultFactory);
-            obj.Services.AddInstance(contributor);
-
-            return obj;
+            return obj.AddContributorNoCheck<T>(contributor);
         }
 
         /// <summary>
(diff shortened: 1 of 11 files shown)


src/Microsoft.Restier.Core/ApiBuilderExtensions.cs
```diff
@@ -381,60 +431,9 @@ public IServiceScope CreateApiScope()
...
-                var paramInfoList = creator.GetParameters();
-                var paramList = new object[paramInfoList.Length];
-                for (int i = 0; i < paramInfoList.Length; i++)
-                {
-                    var paramInfo = paramInfoList[i];
-                    if (paramInfo.ParameterType == serviceType)
-                    {
-                        paramList[i] = next();
-                        continue;
                     }
```
lewischeng-ms: Fix indentation of line 434 and 435.

src/Microsoft.Restier.Core/ApiBuilderExtensions.cs
```diff
@@ -152,11 +152,7 @@ public static ApiBuilder AddContributor<T>(this ApiBuilder obj, ApiServiceContri
             Ensure.NotNull(obj, "obj");
             Ensure.NotNull(contributor, "contributor");
 
-            // Services have singleton lifetime by default, call Make... to change.
-            obj.Services.TryAddSingleton(typeof(T), ChainedService<T>.DefaultFactory);
-            obj.Services.AddInstance(contributor);
-
-            return obj;
+            return obj.AddContributorNoCheck<T>(contributor);
```
lewischeng-ms: Please remove the unused using in this file: `using System.Collections.Generic;`

src/Microsoft.Restier.Core/ApiBuilderExtensions.cs
rayao: Thanks.

src/Microsoft.Restier.Core/ApiBuilderExtensions.cs
rayao: Thanks.

src/Microsoft.Restier.Core/ApiBuilderExtensions.cs
```diff
@@ -186,48 +183,91 @@ public static ApiBuilder ChainPrevious<T>(this ApiBuilder obj, Func<IServiceProv
...
-                var ret = ChainedService.CreateViaReflection(typeof(TService), typeof(TImplement), sp, next);
-                var value = ret.Item1;
-                var creator = ret.Item2;
+                var instance = sp.GetService<TImplement>();
+                if (instance == null)
+                {
+                    return instance;
+                }
+
+                var nextProperty = typeof(TImplement).GetTypeInfo().GetProperties()
```
lewischeng-ms: How about using **non-public field** for "next"? I don't know the answer, just to have some discussion with you.
1) If the next property is public, it probably forces to add a public API to the user code and may be abused (e.g., intentionally set by the user to do something tricky...will make the code hard to maintain)
2) The next handler for a hook handler is more likely to be a private stuff for that handler.
3) If the user really like the field to be public, he can create a public property to return it.

src/Microsoft.Restier.Core/ApiBuilderExtensions.cs
```diff
@@ -186,48 +183,91 @@ public static ApiBuilder ChainPrevious<T>(this ApiBuilder obj, Func<IServiceProv
...
-                // Create a factory proxy for the first call, to save the expression compilation cost to the 2nd
-                // call, in case the service is a singleton or for any reason only resolved once.
-                factory = (serviceProvider, nextService) =>
+                nextProperty.SetValue(instance, next());
+                factory = (serviceProvider, getNext) =>
                 {
-                    factory = ChainedService<TService>.CreateFactory(creator);
-                    creator = null;
-                    return factory(serviceProvider, nextService);
+                    var serviceProviderParam = Expression.Parameter(typeof(IServiceProvider));
```
lewischeng-ms: better add a block comment here to explain what the below chunk of code does. such as:
```
// To build a lambda expression like:
// factory = (sp, next) => 
// {
//     var hook = sp.GetRequiredService<TImplement>();
//     hook.next = next();
//     return hook;
// }
```
Please correct me if my perception is not right:-)

src/Microsoft.Restier.Core/ApiBuilderExtensions.cs
rayao: **Thank you!**
I don't even know it's possible. I supposed that expression compilation needs the property to be public settable. I was an idiot...

src/Microsoft.Restier.Core/ApiBuilderExtensions.cs
```diff
@@ -186,48 +183,91 @@ public static ApiBuilder ChainPrevious<T>(this ApiBuilder obj, Func<IServiceProv
...
+                        Expression.Assign(value, getService),
+                        inject,
+                        value);
+
+                    factory = LambdaExpression.Lambda<Func<IServiceProvider, Func<TService>, TService>>(
+                        block,
+                        serviceProviderParam,
+                        nextParam).Compile();
+                    nextProperty = null;
+                    return factory(serviceProvider, getNext);
```
lewischeng-ms: Then the question comes: can we directly compute the final hook object rather than build a lambda, compile and evaluate? Maybe it's not feasible, I am just guess.

rayao: You mean cache an object rather than call next() to retrieve a new one? That'll break the life time rule of the DI container. The whole purpose of building a lambda is to maintain the life time rule of the registered service.

rayao: Maybe I didn't get your point?

lewischeng-ms: Ohh...my fault...I didn't see factory is a context variable for the lambda...sorry. Ignore me.

src/Microsoft.Restier.Core/ApiBuilderExtensions.cs
rayao: Your comment is precise. :)

src/Microsoft.Restier.Core/ApiBuilderExtensions.cs
lewischeng-ms: Uhh...I thought about this twice. It's struggling for me. Maybe both field and property (with higher property) shall work. But I wonder if we need to set some constraint on the naming? I am really afraid it gives user too much freedom which may cause some problem in the future.

lewischeng-ms:
@rayao I have no other comments for this PR. We can think of the design about how to inject the "next" pointer later. Please merge the PR after fixing the issues. Thanks.

src/Microsoft.Restier.Core/ApiBuilderExtensions.cs
rayao: Naming contraint is kind of "hidden", makes cosumers easier to forget or misspell.
Let's give it a further discussion.

lewischeng-ms: Sure. Current design is practical enough.

rayao: Oh I made it wrong, now it's still property instead of field. Let's correct it in next check in. Sorry.

lewischeng-ms: It's OK. Currently existing hook handlers go through another code path so they seem not to be affected.

PR #297, opened 2016-02-19
DI backed ApiPermission

rayao:
Treat ApiPermission as api services. IMHO the entire role based security system could be registered as several types of api service, but that needs some design discussion. This change is the easiest part.


rayao:
#298

lewischeng-ms:
Looks good to me.

rayao:
Pushed with de696661

PR #302, opened 2016-02-27
Remove IHookHandler and IDelegateHookHandler.

### LINKED ISSUE #289: Remove IDelegateHookHandler<T> and LegacyHookHandler<T> plus the corresponding logic
after we migrate all the existing hook handlers...

### LINKED ISSUE #290: Replace SomeHookHandler.Instance with singleton-lifetime services in DI
such as `DefaultQueryExecutor.Instance`. Since now we have DI, there is no need to maintain so many static singleton instances in various hook handlers, consider replace them with singleton-lifetime services in DI.

rayao:
To fix #289 and resolve #290.
I only removed Instance member for those that I think cache is likely to be added later.


src/Microsoft.Restier.Core/ApiBuilderExtensions.cs
```diff
@@ -73,58 +73,56 @@ public static ApiBuilder TryUseContextApiScope(this ApiBuilder obj)
...
-        /// True if the hook handler is registered.
+        /// True if the service is registered.
         /// </returns>
-        public static bool HasHookHandler<T>(this ApiBuilder obj) where T : class, IHookHandler
+        public static bool HasService<T>(this ApiBuilder obj) where T : class
         {
             Ensure.NotNull(obj, "obj");
 
-            return obj.Services.Any(sd => sd.ServiceType == typeof(LegacyHookHandler<T>));
+            return obj.Services.Any(sd => sd.ServiceType == typeof(ApiServiceContributor<T>));
```
rayao: This should be typeof(T), will change before check in.

src/Microsoft.Restier.Core/Conventions/ConventionBasedApiModelBuilder.cs
```diff
@@ -56,11 +49,14 @@ private ConventionBasedApiModelBuilder(Type targetType)
...
-            var provider = new ConventionBasedApiModelBuilder(targetType);
-            builder.AddHookHandler<IModelBuilder>(provider);
-            builder.AddHookHandler<IModelMapper>(provider);
-            builder.AddHookHandler<IQueryExpressionExpander>(provider);
-            builder.AddHookHandler<IQueryExpressionSourcer>(provider);
+            // The model builder must maintain a singleton life time, for holding states and being injected into
+            // some other services.
+            builder.AddInstance(new ConventionBasedApiModelBuilder(targetType));
+
+            builder.ChainPrevious<IModelBuilder, ConventionBasedApiModelBuilder>();
```
rayao: Will split IModelBuilder to make shared state clearly separated.

src/Microsoft.Restier.Core/Conventions/ConventionBasedApiModelBuilder.cs
```diff
@@ -436,5 +340,131 @@ private void AddNavigationPropertyBindings(IEdmModel model)
                 }
             }
         }
+
+        internal class ModelMapper : IModelMapper
```
rayao: CA can't get a knowledge that this class is actually instantiated, so an error is inevitable.
Also the InnerHandler member for not assigned.
This is kind of harassing, don't know if it could be circumvented.

test/Microsoft.Restier.Core.Tests/Api.Tests.cs
```diff
@@ -18,10 +18,8 @@ namespace Microsoft.Restier.Core.Tests
 {
     public class ApiTests
     {
-        private class TestModelBuilder : IDelegateHookHandler<IModelBuilder>, IModelBuilder
+        private class TestModelBuilder : IModelBuilder
         {
-            public IModelBuilder InnerHandler { get; set; }
```
rayao: Seems this is not actually used.

src/Microsoft.Restier.Security/ApiConfigurationExtensions.cs
```diff
@@ -32,8 +32,8 @@ public static class ApiConfigurationExtensions
             this ApiBuilder builder)
         {
             Ensure.NotNull(builder, "builder");
-            builder.AddHookHandler<IQueryExpressionInspector>(RoleBasedAuthorization.Default);
-            builder.AddHookHandler<IQueryExpressionExpander>(RoleBasedAuthorization.Default);
+            builder.CutoffPrevious<IQueryExpressionInspector, RoleBasedAuthorization>();
+            builder.ChainPrevious<IQueryExpressionExpander, RoleBasedAuthorization>();
```
rayao: This seems to be a severe bug that sematic changed with life time change.

rayao: Since RoleBasedAuthorization has no state, this change won't break anything. But still I'd like to split the 2 services to make it clearer and avoid any inadvertent change to break the semantic.

src/Microsoft.Restier.Core/Conventions/ConventionBasedApiModelBuilder.cs
```diff
@@ -56,130 +46,14 @@ private ConventionBasedApiModelBuilder(Type targetType)
...
-            var query = GetEntitySetQuery(context) ?? GetSingletonQuery(context);
-            if (query != null)
-            {
-                return Expression.Constant(query);
-            }
-
-            return null;
+            builder.ChainPrevious<IModelBuilder, ModelBuilder>();
+            builder.ChainPrevious<IModelMapper, ModelMapper>();
+            builder.CutoffPrevious<IQueryExpressionExpander, QueryExpressionExpander>();
```
rayao: Can we safely assume this is the first registered IQueryExpressionExpander?

src/Microsoft.Restier.Security/RoleBasedAuthorization.cs
```diff
@@ -17,28 +17,10 @@
 namespace Microsoft.Restier.Security
 {
     /// <summary>
-    /// Represents a role-based authorization system.
+    /// This class applies security policy on expanded expression nodes.
     /// </summary>
-    public class RoleBasedAuthorization : IQueryExpressionInspector,
-        IQueryExpressionExpander, IDelegateHookHandler<IQueryExpressionExpander>
+    public class ApiPolicyActivator : IQueryExpressionExpander
```
rayao: Let me know if you want another name.

rayao:
Also resolve #289 after commit.

rayao:
@lewischeng-ms About naming, I suggest 3 options:

- The code in the pull request now, **CutoffPrevious** for ignoring all previously registered. **ChainPrevious** for chaining previously registered.
- **SetHookPoint** (or maybe **SetApiService**?) for ignoring all previously registered, **AddHookPoint** for chaining previous. This is to make it familiar to early consumers.
- **ApiBuilder.ChainedService<T>().Cutoff** for ignoring previous, **ChainedService<T>().Chain** for chaining previous. This is to make it clear that the service is chained, and since we now have regular services supported by DI, perhaps this API level discern is developer friendly.

src/Microsoft.Restier.Core/Conventions/ConventionBasedApiModelBuilder.cs
lewischeng-ms: I think so. All convention-based or default hook handlers should be the inner most.

lewischeng-ms:
@rayao, Sorry to review your PR so late! For now I think we don't need to make a final call on the API. `CutoffPrevious` and `ChainPrevious` have very clear meaning thus are good enough for now. But I am curious about 3rd option you suggested. What is `ChainedService()` going to return? Or did you mean `ChainedService<T>()` to return the existing chained services?

As always thanks for your great help on this!

test/Microsoft.Restier.Core.Tests/Api.Tests.cs
lewischeng-ms: Yes, maybe it's been added by mistake. Please feel free to remove it.

src/Microsoft.Restier.Core/Conventions/ConventionBasedApiModelBuilder.cs
lewischeng-ms: Do you mean to split the two functionalities (`IModelBuilder` and `IModelMapper`) of `ConventionBasedApiModelBuilder`?

src/Microsoft.Restier.Core/Conventions/ConventionBasedApiModelBuilder.cs
lewischeng-ms: Please feel free to suppress such CA errors.

src/Microsoft.Restier.Security/ApiConfigurationExtensions.cs
lewischeng-ms: I agree. There were so many classes in RESTier that took multiple duties (hook handlers). That makes the logic unclear. IMHO even if we have something to share among the "logically" two hook handlers, we shall seek another way to achieve that (such as putting shared state into DI). Does it make sense?

src/Microsoft.Restier.Security/RoleBasedAuthorization.cs
lewischeng-ms: I think it's accurate to describe the class here. Please go ahead. By the way, to keep consistency, I think `AssertAttribute` should be renamed to `AssertRoleAttribute` and `RoleBasedAuthorization` should be renamed to `RoleBasedAuthorizer`. But that's OK if you feel this should be a separate task.

rayao: Ok，I'll include them in this change.

src/Microsoft.Restier.Security/ApiConfigurationExtensions.cs
rayao: > even if we have something to share among the "logically" two hook handlers, we shall seek another way to achieve that (such as putting shared state into DI). 
This done **EXACTLY** in this change, to class ConventionBasedApiModelBuilder. :)

rayao: > even if we have something to share among the "logically" two hook handlers, we shall seek another way to achieve that (such as putting shared state into DI). 

This is done **EXACTLY** in this change, to ConventionBasedApiModelBuilder. :)

rayao:
@lewischeng-ms Yeah exactly as your thought. I just didn't realize 'gt' and 'lt' are omitted without even notifying me.
No problem, as always please just go at your own pace, the least I want is that you felt urged.

src/Microsoft.Restier.Core/Conventions/ConventionBasedApiModelBuilder.cs
rayao: What I worried about is that `ApiBase.ConfigureApi` is called before `EnableConventions`, which means if a data provider registers IQueryExpressionExpander, this call will cut off that registration.

lewischeng-ms: Sorry I made a mistake. In fact we have to distinguish between default and convention-based handlers where default handlers should be the innermost while the convention-based ones should be the outermost.

rayao: Then shall I change it to ChainPrevious?

src/Microsoft.Restier.Security/ApiConfigurationExtensions.cs
lewischeng-ms: Awesome!

src/Microsoft.Restier.Security/RoleBasedAuthorization.cs
lewischeng-ms: Please also rename the `RoleBaseAuthorization.cs`. Thanks.

lewischeng-ms:
@rayao I've no other comments on this PR. Please fix the name of `RoleBasedAuthorization.cs` and merge. Thanks.

src/Microsoft.Restier.Core/Conventions/ConventionBasedApiModelBuilder.cs
lewischeng-ms: Yeah, I think so.

PR #306, opened 2016-03-07
Thread-safe GetModelAsync

### LINKED ISSUE #304: If GetModelAsync takes too long to complete, requests will fail.
Since GetModelAsync is asynchronous, if it takes too long, RESTier routing won't get registered in time and all requests before the registration done will fail.
Also, in test runs sometimes 2 or more test methods run simultaneously, which leads to GetModelAsync called simultaneously for a same Api, the model pipeline will throw for lacking synchronization mechanism. I think this could be triggered in production when 2 services expose a same Api.

rayao:
This ONLY synchronizes multiple simultaneous calls to model building pipeline, WebApiConfig still could return before a long running GetModelAsync completes.
Partially resolve #304

### DIFF
--- a/src/Microsoft.Restier.Core/Api.cs
+++ b/src/Microsoft.Restier.Core/Api.cs
@@ -78,16 +78,40 @@ namespace Microsoft.Restier.Core
         {
             Ensure.NotNull(context, "context");
             var config = context.Configuration;
-            if (config.Model == null)
+            if (config.Model != null)
             {
-                var builder = context.GetApiService<IModelBuilder>();
-                if (builder != null)
-                {
-                    config.Model = await builder.GetModelAsync(new InvocationContext(context), cancellationToken);
-                }
+                return config.Model;
             }
 
-            return config.Model;
+            var builder = context.GetApiService<IModelBuilder>();
+            if (builder == null)
+            {
+                return null;
+            }
+
+            Task<IEdmModel> running;
+            var source = config.CompeteModelGeneration(out running);
+            if (source == null)
+            {
+                return await running;
+            }
+
+            try
+            {
+                var model = await builder.GetModelAsync(new InvocationContext(context), cancellationToken);
+                source.SetResult(model);
+                return model;
+            }
+            catch (AggregateException e)
+            {
+                source.SetException(e.InnerExceptions);
+                throw;
+            }
+            catch (Exception e)
+            {
+                source.SetException(e);
+                throw;
+            }
         }
 
         #endregion
--- a/src/Microsoft.Restier.Core/Api.cs
+++ b/src/Microsoft.Restier.Core/Api.cs
@@ -86,7 +86,7 @@ namespace Microsoft.Restier.Core
             var builder = context.GetApiService<IModelBuilder>();
             if (builder == null)
             {
-                return null;
+                throw new InvalidOperationException(Resources.ModelBuilderNotRegistered);
             }
 
             Task<IEdmModel> running;
--- a/src/Microsoft.Restier.Core/ApiConfiguration.cs
+++ b/src/Microsoft.Restier.Core/ApiConfiguration.cs
@@ -2,9 +2,10 @@
 // Licensed under the MIT License.  See License.txt in the project root for license information.
 
 using System;
+using System.Threading;
+using System.Threading.Tasks;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.OData.Edm;
-using Microsoft.Restier.Core.Properties;
 
 namespace Microsoft.Restier.Core
 {
@@ -35,6 +36,8 @@ namespace Microsoft.Restier.Core
     {
         private IServiceProvider serviceProvider;
 
+        private Task<IEdmModel> modelTask;
+
         /// <summary>
         /// Initializes a new instance of the <see cref="ApiConfiguration" /> class.
         /// </summary>
@@ -54,7 +57,7 @@ namespace Microsoft.Restier.Core
             get { return serviceProvider; }
         }
 
-        internal IEdmModel Model { get; set; }
+        internal IEdmModel Model { get; private set; }
 
         /// <summary>
         /// Gets a service instance.
@@ -64,6 +67,47 @@ namespace Microsoft.Restier.Core
         public T GetApiService<T>() where T : class
         {
             return this.serviceProvider.GetService<T>();
+        }
+
+        internal TaskCompletionSource<IEdmModel> CompeteModelGeneration(out Task<IEdmModel> running)
+        {
+            var source = new TaskCompletionSource<IEdmModel>(TaskCreationOptions.AttachedToParent);
+            var exist = Interlocked.CompareExchange(ref modelTask, source.Task, null);
+            if (exist != null)
+            {
+                running = exist;
+                source.SetCanceled();
+                return null;
+            }
+
+            source.Task.ContinueWith(
+                task => FinishModelGeneration(task),
+                TaskContinuationOptions.ExecuteSynchronously);
+            running = null;
+            return source;
+        }
+
+        private void FinishModelGeneration(Task<IEdmModel> task)
+        {
+            if (task.Status == TaskStatus.RanToCompletion)
+            {
+                ReplaceModelTask(Task.FromResult(task.Result), task);
+                Model = task.Result;
+            }
+            else
+            {
+                // Set modelTask null to allow retrying GetModelAsync.
+                ReplaceModelTask(null, task);
+            }
+        }
+
+        private void ReplaceModelTask(Task<IEdmModel> value, Task<IEdmModel> comparand)
+        {
+            var exist = Interlocked.CompareExchange(ref modelTask, value, comparand);
+            if (exist != comparand)
+            {
+                throw new InvalidOperationException("Unexpected: model task mismatch");
+            }
         }
     }
 }
--- a/src/Microsoft.Restier.Core/ApiConfiguration.cs
+++ b/src/Microsoft.Restier.Core/ApiConfiguration.cs
@@ -72,42 +72,30 @@ namespace Microsoft.Restier.Core
         internal TaskCompletionSource<IEdmModel> CompeteModelGeneration(out Task<IEdmModel> running)
         {
             var source = new TaskCompletionSource<IEdmModel>(TaskCreationOptions.AttachedToParent);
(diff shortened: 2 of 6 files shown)


src/Microsoft.Restier.Core/Api.cs
```diff
@@ -78,16 +78,40 @@ public static class Api // TODO GitHubIssue#25,#26 : transactions, exception fil
...
-                {
-                    config.Model = await builder.GetModelAsync(new InvocationContext(context), cancellationToken);
-                }
+                return config.Model;
+            }
+
+            var builder = context.GetApiService<IModelBuilder>();
+            if (builder == null)
+            {
+                return null;
```
lewischeng-ms: Maybe we should throw if there is no model builder? It is more meaningful than a later null-reference exception.

src/Microsoft.Restier.Core/ApiConfiguration.cs
```diff
@@ -65,5 +68,46 @@ public IServiceProvider ServiceProvider
         {
             return this.serviceProvider.GetService<T>();
         }
+
+        internal TaskCompletionSource<IEdmModel> CompeteModelGeneration(out Task<IEdmModel> running)
```
lewischeng-ms: Compete or Complete?

lewischeng-ms:
So basically this PR is to make any call to `GetModelAsync` to wait for the first one to complete? If we need to sync with the later requests, we have to register a model factory to the underlying OData Web API rather than the model instance itself. Is my understanding correct?

src/Microsoft.Restier.Core/ApiConfiguration.cs
lewischeng-ms: OK I got that. This should be Compete.

src/Microsoft.Restier.Core/Api.cs
```diff
@@ -78,16 +78,40 @@ public static class Api // TODO GitHubIssue#25,#26 : transactions, exception fil
...
             }
 
-            return config.Model;
+            try
+            {
+                var model = await builder.GetModelAsync(new InvocationContext(context), cancellationToken);
+                source.SetResult(model);
+                return model;
+            }
+            catch (AggregateException e)
```
lewischeng-ms: Are we going to do anything special for an `AggregateException` in the future? If not, we may consider merge the two catch-clauses.

src/Microsoft.Restier.Core/ApiConfiguration.cs
```diff
@@ -2,9 +2,10 @@
 // Licensed under the MIT License.  See License.txt in the project root for license information.
 
 using System;
+using System.Threading;
+using System.Threading.Tasks;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.OData.Edm;
-using Microsoft.Restier.Core.Properties;
```
lewischeng-ms: Thanks for cleaning up unused namespaces!

src/Microsoft.Restier.Core/ApiConfiguration.cs
```diff
@@ -65,5 +68,46 @@ public IServiceProvider ServiceProvider
...
+            var source = new TaskCompletionSource<IEdmModel>(TaskCreationOptions.AttachedToParent);
+            var exist = Interlocked.CompareExchange(ref modelTask, source.Task, null);
+            if (exist != null)
+            {
+                running = exist;
+                source.SetCanceled();
+                return null;
+            }
+
+            source.Task.ContinueWith(
```
lewischeng-ms: A naïve question: shall we set the returned task to `source.Task` otherwise the `ContinueWith` part will be lost?

src/Microsoft.Restier.Core/ApiConfiguration.cs
```diff
@@ -65,5 +68,46 @@ public IServiceProvider ServiceProvider
         {
             return this.serviceProvider.GetService<T>();
         }
+
+        internal TaskCompletionSource<IEdmModel> CompeteModelGeneration(out Task<IEdmModel> running)
+        {
+            var source = new TaskCompletionSource<IEdmModel>(TaskCreationOptions.AttachedToParent);
+            var exist = Interlocked.CompareExchange(ref modelTask, source.Task, null);
```
lewischeng-ms: Shall we name it `runningTask` to make the code clearer? For example, if we wrote `if (runningTask != null)`, we knew that there is no ongoing task.

src/Microsoft.Restier.Core/ApiConfiguration.cs
```diff
@@ -65,5 +68,46 @@ public IServiceProvider ServiceProvider
...
+                TaskContinuationOptions.ExecuteSynchronously);
+            running = null;
+            return source;
+        }
+
+        private void FinishModelGeneration(Task<IEdmModel> task)
+        {
+            if (task.Status == TaskStatus.RanToCompletion)
+            {
+                ReplaceModelTask(Task.FromResult(task.Result), task);
```
lewischeng-ms: So this is to replace the model task with a straightforward task which directly returns the model instance?

src/Microsoft.Restier.Core/Api.cs
rayao: Ok, thanks.

src/Microsoft.Restier.Core/ApiConfiguration.cs
rayao: You mean set the out running parameter? It's not necessary I think, since source.Task continues when SetResult or SetException is called in GetModelAsync, that will trigger the continuation here.

src/Microsoft.Restier.Core/ApiConfiguration.cs
rayao: Partially, the main purpose is to confirm that this is truly the original model generation task. But actually it must be. In a previous version FinishModelGeneration is called from GetModelAsync, so I added this validation.
And I didn't look into Task code but intuitively I think a task built from direct result would consume less resources than a completed task from a completion source.

src/Microsoft.Restier.Core/Api.cs
lewischeng-ms: That sounds good:-) Thanks.

src/Microsoft.Restier.Core/ApiConfiguration.cs
lewischeng-ms: Sorry I mean `Task.ContinueWith` returns a continuation task which will be called when, as you said, `SetResult` or `SetException` is called. But does it have a side effect that the original task is appended with the continued task as well? If not, `source.Task` will return the original one rather than the continued one. But that may not be a problem though.

src/Microsoft.Restier.Core/ApiConfiguration.cs
lewischeng-ms: Intuitively I think a `Task` instance primarily contains the id information, state variable, the pointer to a task code. So I would assume that all task instances are of the same size. So it's OK to leave the completed task as is. But I do think the design to replace the `modelTask` with `null` to allow retry is very good:)

src/Microsoft.Restier.Core/ApiConfiguration.cs
rayao: Thanks.

src/Microsoft.Restier.Core/Api.cs
rayao: If we catch AggregatedException as normal Exception, and SetException to the task, the task will wrap it again and so ends with an AggreatedException, which has an inner AggregatedException, which has the actual exceptions as its InnerExceptions.

src/Microsoft.Restier.Core/ApiConfiguration.cs
rayao: Oh yeah, I got your thought. Yes another thread would continue on the original task source task, not the appended one, so it could possibly run before EdmModel is set, but as you pointed out this won't hurt, since anyway the resulting IEdmModel is retrieved.

src/Microsoft.Restier.Core/ApiConfiguration.cs
rayao: Removed the replace call and adds an assert.
`Debug.Assert(task == modelTask);`
Thanks.

rayao:
> So basically this PR is to make any call to  GetModelAsync  to wait for the first one to complete? If we need to sync with the later requests, we have to register a model factory to the underlying OData Web API rather than the model instance itself. Is my understanding correct?
Perhaps I don't get your idea clearly, since GetModelAsync works on ApiConfiguration, so calling it with any model instance will be synchronized. So, if a query needs EdmModel to complete, an `await GetModelAsync` will guarantee that.
The real problem is the async nature of GetModelAsync, since WebApiConfig.Register is synchronous, it can't wait Restier route to be completely established, for example:
```
        public static async void RegisterNorthwind(
            HttpConfiguration config, HttpServer server)
        {
            await config.MapRestierRoute<NorthwindApi>(
                "NorthwindApi", "api/Northwind",
                new RestierBatchHandler(server));
        }
```
RegisterNorthwind could have returned before the route is completely registered. So if the model generation is really costly, RESTier routing will likely get messed up.
You can put an `await Task.Delay(5000)` before the config.Map call to repro.

lewischeng-ms:
@rayao Yeah, I understand your concern about requests coming before the route has been completely registered. But in this case, since the route has not been registered yet. The service could safely return 404 for those requests. And what I was thinking about was that you can't process any OData request without getting the model instance (parse URI or serialization/deserialization). Now the model instance is registered staticly into `ODataPathRouteConstraint`. If we could register a factory method to produce the model into Web API and let the factory method call `await GetModelAsync()`, the service could probably wait for model build completion then process the first incoming request.

lewischeng-ms:
I've no other questions, please feel free to merge it.

rayao:
> But in this case, since the route has not been registered yet. The service could safely return 404 for those requests

@lewischeng-ms This is only true for synchronous long running GetModelAsync call (like all of our existing samples, although they return task but they're synchronous in nature). A truly async GetModelAsync call won't have the chance to registered a route before WebApiConfig pipeline ends, and that results in RESTier routing not established, **no further requests could be processed.**
You can add an `await Task.Delay(1000)` in model pipeline to see the effect.

> If we could register a factory method to produce the model into Web API and let the factory method call  await GetModelAsync() 

Yeah I initially thought that, but since the route selection pipeline is synchronous, we can't await.
I've checked in this change and working on another PR which could possibly be the resolution.

PR #313, opened 2016-03-07
Fix various design issues in RESTier

### LINKED ISSUE #308: Fix NullReferenceException when ColumnTypeAttribute does not have a TypeName property specified

### LINKED ISSUE #309: Fix QueryExpressionSourcer in EF provider to ignore entity sets that don't come from DbSet properties

### LINKED ISSUE #310: Fix the behavior of QueryExpressionExpander to correctly expand the entity set properties

### LINKED ISSUE #311: Make all the convention-based handlers outermost
And let those handlers call custom handlers first. In this way, we don't need to worry about applying conventions in the wrong order.

### LINKED ISSUE #312: QueryExpressionVisitor.BaseQuery is not set when the query is not generated by ApiData
Regardless whether the query is generated by ApiData or not (which means directly returned by a property), the BaseQuery needs to be set anyway. If it's the second case, the BaseQuery should be set to the value of the first constant expression (not changed by any hook handler) visited.

lewischeng-ms:
https://github.com/lewischeng-ms/RESTier/commit/6256eaf3b67f80302c77f168b4427f78fcb565c1 resolves #308.
https://github.com/lewischeng-ms/RESTier/commit/654ac34a7c15b64d8afd7105123c87f20c264be5 resolves #309.
https://github.com/lewischeng-ms/RESTier/commit/fec87c969357f22536d69c8fb0405106e4c176ec resolves #310.
https://github.com/lewischeng-ms/RESTier/commit/ad2d882abb1aae4050e32905af216c333d0fa0ef resolves #311.
https://github.com/lewischeng-ms/RESTier/commit/d326e780bc6c404b3236b14a3e094d71ab67cf8d resolves #312.


src/Microsoft.Restier.Security/ApiPolicyActivator.cs
```diff
@@ -33,15 +33,6 @@ public Expression Expand(QueryExpressionContext context)
...
-                var result = this.InnerHandler.Expand(context);
-                if (result != null)
-                {
-                    return result;
-                }
-            }
-
             if (context.ModelReference == null)
             {
                 return null;
```
rayao: Maybe every `return null` should be `return CallInner()`

lewischeng-ms: Yeah, you are right. It MAY have inner expander from the provider to call. Thanks for the reminder.

PR #315, opened 2016-03-08
Fix CA  errors in nightly build.

rayao:
Don't why the suppression didn't do its work. Turn to safest way.


rayao:
And only nightly failed, rolling passed, why?
If nobody gives comments I'll check in before this afternoon.

lewischeng-ms:
That's because nightly is based on Release configuration. Can we simply ignore those CA errors? I want to preserve the code you wrote:-)

rayao:
Thank you! :)
Then I'll build with release and try suppressing.

PR #316, opened 2016-03-08
Fix CA errors, try suppress them under release build.

rayao:


lewischeng-ms:
There is no issue under release on TFS. So we actually don't need to fix any code. Please close this PR. Thanks.

rayao:
Thank you for looking into this.

PR #317, opened 2016-03-08
Scoped Api

### LINKED ISSUE #298: Should we enforce contextual service scope?mode
Currently an Api class could choose between shared and contextual service scope, that makes it impossible to unleash the power of DI, especially stateful service. Maybe we'd enforce the contextual service scope.
For example, if a contextual scope is assured, we can register DbContext as a service, in DbApi.

rayao:
Make both IApi and ApiContext scoped services.
Resolves #298 

Make DbContext scoped.


src/Microsoft.Restier.Core/ApiBase.cs
```diff
@@ -112,7 +124,12 @@ protected ApiContext ApiContext
         /// </summary>
         public void Dispose()
         {
-            if (!this.IsDisposed && this.apiContext != null)
+            if (this.IsDisposed)
```
lewischeng-ms: Thanks for fixing the bug!

src/Microsoft.Restier.Core/ApiBase.cs
```diff
@@ -173,15 +190,11 @@ protected virtual ApiConfiguration CreateApiConfiguration(ApiBuilder builder)
         /// </param>
         protected virtual void Dispose(bool disposing)
         {
+            this.IsDisposed = true;
```
lewischeng-ms: Shall we put `this.IsDisposed` to after the scope has been disposed.

rayao: Disposing the scope will get every service in the scope to be disposed, including this (IApi). So actually `ApiBase.Dispose()` will get a recursive call from ServiceScope, put the assignment before disposing the scope prevents `ApiConfiguratorAttribute.ApplyDisposal` called twice.

lewischeng-ms: I see. Thanks.

src/Microsoft.Restier.Core/ApiBase.cs
```diff
@@ -214,5 +227,13 @@ protected virtual void Dispose(bool disposing)
             ConventionBasedOperationProvider.ApplyTo(builder, targetType);
             ConventionBasedEntitySetFilter.ApplyTo(builder, targetType);
         }
+
+        // Registered as a scoped service so that IApi and ApiContext could be exposed as scoped service.
+        // If a descendant class wants to expose these 2 services in another way, it must ensure they could be
+        // resolved after CreateApiContext call.
+        private class ApiScope
```
lewischeng-ms: Previously I thought this was the scope class but I found it doesn't inherit `IServiceScope` and its only usage is to keep the instance of an `Api` class which will be later used in service factory methods. So I suggest we use a different name for this class like `ApiKeeper`, `ApiHolder`, or else to avoid any confusion with the context scope.

src/Microsoft.Restier.Core/ApiBase.cs
```diff
@@ -67,7 +67,14 @@ protected ApiConfiguration ApiConfiguration
                         var builder = this.ConfigureApi(new ApiBuilder());
                         EnableConventions(builder, apiType);
                         ApiConfiguratorAttribute.ApplyApiBuilder(apiType, builder);
-                        builder.TryUseSharedApiScope(); // TODO: Maybe default to context scope?
+                        if (!builder.HasService<IApi>())
+                        {
+                            builder.Services
+                                .AddScoped(apiType, sp => sp.GetService<ApiScope>().Api)
+                                .AddScoped<IApi>(sp => sp.GetService<ApiScope>().Api)
+                                .AddScoped<ApiContext>(sp => sp.GetService<ApiScope>().Api.ApiContext)
```
lewischeng-ms: I think we can eliminate the generic parameter since the return type of the lambda is `ApiContext`.

src/Microsoft.Restier.Core/ApiBase.cs
```diff
@@ -67,7 +67,14 @@ protected ApiConfiguration ApiConfiguration
...
+                                .AddScoped(sp => new ApiScope());
```
lewischeng-ms: Can we write it as `.AddScoped<ApiScope>())`?

src/Microsoft.Restier.Core/Conventions/ConventionBasedApiModelBuilder.cs
```diff
@@ -137,8 +136,7 @@ private IQueryable GetSingletonQuery(QueryExpressionContext context)
                 object target = null;
                 if (!singletonProperty.GetMethod.IsStatic)
                 {
-                    target = context.QueryContext.ApiContext
-                        .GetProperty(typeof(Api).AssemblyQualifiedName);
+                    target = context.QueryContext.GetApiService<IApi>();
```
lewischeng-ms: Just for design discussion: Since we already have an abstract class called `ApiBase` which is used as the base class for all the `Api` classes (we don't expect the user to directly implement an `Api` class from `IApi`), do you think we should still keep `IApi`? I didn't see its value here.

src/Microsoft.Restier.Core/ApiBase.cs
```diff
@@ -67,7 +67,14 @@ protected ApiConfiguration ApiConfiguration
                         var builder = this.ConfigureApi(new ApiBuilder());
```
lewischeng-ms: A side question here: I saw there is a class called `ContainerBuilder` in https://docs.asp.net/en/latest/fundamentals/dependency-injection.html?highlight=dependency%20injection#replacing-the-default-services-container. But I cannot find it in the latest source of DI. I wonder if the `IServiceCollection` and `ServiceCollection` now play the role of container here? Thanks.

src/Microsoft.Restier.Core/ApiBase.cs
```diff
@@ -67,7 +67,14 @@ protected ApiConfiguration ApiConfiguration
...
                         EnableConventions(builder, apiType);
                         ApiConfiguratorAttribute.ApplyApiBuilder(apiType, builder);
-                        builder.TryUseSharedApiScope(); // TODO: Maybe default to context scope?
+                        if (!builder.HasService<IApi>())
+                        {
+                            builder.Services
+                                .AddScoped(apiType, sp => sp.GetService<ApiScope>().Api)
```
lewischeng-ms: Shall we just use `GetService<IApi>` so here you don't have to register the service for `apiType`?

src/Microsoft.Restier.EntityFramework/DbApi.cs
```diff
@@ -56,71 +57,40 @@ protected T DbContext
...
-            var dbContext = this.CreateDbContext();
+            builder = base.ConfigureApi(builder)
+                .CutoffPrevious<IModelBuilder>(ModelProducer.Instance)
+                .CutoffPrevious<IModelMapper>(new ModelMapper(typeof(T)))
+                .CutoffPrevious<IQueryExpressionSourcer, QueryExpressionSourcer>()
+                .CutoffPrevious<IQueryExecutor>(QueryExecutor.Instance)
+                .CutoffPrevious<IChangeSetPreparer, ChangeSetPreparer>()
+                .CutoffPrevious<ISubmitExecutor>(SubmitExecutor.Instance);
+            builder.Services
+                .AddScoped<T>(sp =>
```
lewischeng-ms: The same question here: Shall we just use `GetService<DbContext>` so here you don't have to register the service for `T` (the actual type of `DbContext`)?

lewischeng-ms:
@rayao Overall, this PR is amazing! It makes the API more compact and beautiful. Thanks for your contribution!

src/Microsoft.Restier.Core/ApiBase.cs
rayao: That's from Autofac, https://github.com/autofac/Autofac. And I also think it acts as the ServiceCollection in MS.DI
And actually I have another topic want to discuss with you.
Initially I wanted to expose IServiceCollection directly for service registration, and then for some reasons added a wrapper class ApiBuilder.
- AddContributor/ChainPrevious APIs, they're somehow unique to RESTier. But actually we could also let them to be extension methods of IServiceCollection.
- Some default services and behaviors in ApiBuilder, but those could also be taken care of with some extension methods.
- Basic service registration could be fulfilled by ApiBuilder, need not a reference to MS.DI, but this only true for simple (and most likely RESTier specific) usage of DI.
- ApiBuilder somehow hides CLSCompliant(false).

Do you prefer IServiceCollection directly or a wrapper like ApiBuilder? Sorry for biinging this up so late, if you like the other way I'd like to take the tedious work. It is said that "Those who pollute should clean up." :)

src/Microsoft.Restier.Core/Conventions/ConventionBasedApiModelBuilder.cs
rayao: Personally I incline to keep a pure abstraction, but if you do think we should prevent another base implementation of Api then I agree with you.

src/Microsoft.Restier.EntityFramework/DbApi.cs
```diff
@@ -56,71 +57,40 @@ protected T DbContext
...
+                {
+                    var dbContext = this.CreateDbContext(sp);
```
rayao: Bug here, should be `sp.GetService<IApi>().CreateDbContext`

src/Microsoft.Restier.Core/ApiBase.cs
rayao: This is to ease those custom services that written only for a specific API type, so that they can GetApiService<MyApi>() and access its unique members directly, or even inject MyApi to a service's ctor. Otherwise they'd have to cast first.
I know this scenario is rare, maybe doesn't worth special care, let me remove them if you think unnecessary.

src/Microsoft.Restier.EntityFramework/DbApi.cs
rayao: This is also to ease those service that already know in which Api type they're used. If you think unnecessary I'll remove it.
And actually I want another design discussion here.
Should we just register T normally to use its constructor to instantiate? And of course CreateDbContext should be removed.
- Subclass could still override by registering T in ConfigApi.
- The problem is `ProxyCreationEnabled = false`. Maybe we can leave that to developer?

src/Microsoft.Restier.Core/Conventions/ConventionBasedApiModelBuilder.cs
lewischeng-ms: OK, I see. Let's discuss it later. Thanks for your insight:-)

src/Microsoft.Restier.Core/ApiBase.cs
lewischeng-ms: No worries after all RESTier is still in its preview stage:-) In my personal opinion:

 - I have no objection to both ways: `IServiceCollection` makes API compact while `ApiBuilder` offers a clearer meaning.
 - I just noticed that all methods of `ApiBuilder` are placed in `ApiBuilderExtensions`. Since that we have full control of the implementation of `ApiBuilder`, that `ApiBuilder` doesn't implement any interface, and that methods in `ApiBuilderExtensions` do not rely on pure abstraction either, then why do we use extension methods instead of encapsulation? Just in this sense, I would prefer `IServiceCollection`.

Please correct me if my understanding is wrong. By the way, it is not "pollution", it is just a brilliant idea that doesn't come true:-)

src/Microsoft.Restier.Core/ApiBase.cs
lewischeng-ms: You win:-) I think your comment makes more sense. And it doesn't hurt anything to keep it. The same for `DbContext` below. Thanks for your clarification.

src/Microsoft.Restier.EntityFramework/DbApi.cs
lewischeng-ms: Yeah, this is also an alarm that we should carefully use `this` in a lambda.

src/Microsoft.Restier.EntityFramework/DbApi.cs
lewischeng-ms: Good idea!

lewischeng-ms:
@rayao Please feel free to merge whenever you think this PR is ready. Thanks.

src/Microsoft.Restier.Core/Conventions/ConventionBasedApiModelBuilder.cs
rayao: So let's keep IApi for now, and later decide its future.

src/Microsoft.Restier.Core/ApiBase.cs
rayao: Then let's leave it as now and make decision on ApiBuilder later. At least for now it helps reduce usage of CLSCompliant.

rayao:
Resolved comments and pushed with e88304b1eaefa353c292b68cbc3ea8666ab6eccb
Added a `DbApiBase<T>` class, to keep suppressing proxy creation under EF6. If there's concern let's discuss and change it later.

lewischeng-ms:
@rayao Do we necessarily have to add the `DbApiBase<T>` class? Are we going to add something like `DbApi7<T>` in the future? Even if so, personally I would suggest we keep the API simple for now and it won't be breaking change when the time comes that we have to introduce the `DbApiBase` for large part of common logic between EF6 and EF7.

rayao:
It's not for EF6 or EF7, I just want to keep the proxy suppression behavior in DbApi, if we register DbContext with `AddScope<T>`, we have no chance to suppress proxy creation. So in DbApi, T is created with new(), then set the suppression property.

lewischeng-ms:
Based on our offline discussion, I will change `DbApi` back for now. Thanks.

rayao:
Sure, thank you!

PR #320, opened 2016-03-11
Design Proposal: Run count query independently

### LINKED ISSUE #319: Discussion: Should be run count query independently?
IMHO the reason that current query pipeline includes a totalcount in result, is purely for corresponding to OData/WebApi query options. But the query pipeline should/could and actually is decoupled from OData, such correspondent is unnecessary and when present (as for now), brings about unnecessary complexity to RESTier core.
The core has to reverse engineering the expression to get a count query, this reverse engineering process actually assumes the query is from AspNet.WebApi. That hurts the cohesion of RESTier core.
By separating count query at publisher layer, the simplicity of core is ensured, and publisher still keeps the flexibility.

rayao:
I don't want to check in this, just to show its simplicity.

Pro:
- Release the core from complex count query calculation.
- Make the concern of Core and Publisher more clearly separated.

Con:
- A little perf hit from the extra query building. This could be mitigated by running the 2 queries simultaneously, which is in this review.
- 20+ UTs breaking.

Resolves #319


rayao:
According to #323 this is not the solution we're pursuing.

PR #322, opened 2016-03-14
Make a truly async model pipeline work for WebApi publisher.

rayao:
A truly async model pipeline leads to unsuccessful routing registration. We didn't see this in our existing samples because their model pipelines are synchronous in nature.
This change tries to defer the actual RESTier routing setup until the model pipeline returns. Before that, requests run on a placeholder routing.

### DIFF
--- a/src/Microsoft.Restier.WebApi/HttpConfigurationExtensions.cs
+++ b/src/Microsoft.Restier.WebApi/HttpConfigurationExtensions.cs
@@ -14,6 +14,9 @@ using Microsoft.OData.Edm;
 using Microsoft.Restier.Core;
 using Microsoft.Restier.WebApi.Batch;
 using Microsoft.Restier.WebApi.Routing;
+using System.Web.Http.Routing;
+using System.Net.Http;
+using Microsoft.OData.Edm.Library;
 
 namespace Microsoft.Restier.WebApi
 {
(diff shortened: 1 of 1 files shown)


rayao:
#304

src/Microsoft.Restier.WebApi/HttpConfigurationExtensions.cs
```diff
@@ -44,20 +47,24 @@ public static class HttpConfigurationExtensions
...
+                    ODataRouteConstants.Batch : (routePrefix + '/' + ODataRouteConstants.Batch);
+                routes.MapHttpBatchRoute(routeName + "Batch", routeTemplate, batchHandler);
+            }
 
-                if (batchHandler != null && batchHandler.ApiFactory == null)
-                {
-                    batchHandler.ApiFactory = apiFactory;
-                }
+            var pathHandler = new DefaultODataPathHandler();
+            ////defaultODataPathHandler.ResolverSetttings = config.GetResolverSettings();
```
rayao: ResolverSettings is an internal member, must find a way to set it if no other convenient API.

src/Microsoft.Restier.WebApi/HttpConfigurationExtensions.cs
```diff
@@ -115,5 +122,79 @@ private static void EnsurePayloadValueConverter(this IEdmModel model)
...
+            public Func<IApi> ApiFactory { get; private set; }
+
+            public IEdmModel FinalModel { get; private set; }
+
+            public RestierRouteConstraint(IODataPathHandler pathHandler, string routeName, Func<IApi> apiFactory)
+            {
+                var conventions = ODataRoutingConventions.CreateDefault();
+                Placeholder = new ODataPathRouteConstraint(pathHandler, new EdmModel(), routeName, conventions);
+                ApiFactory = apiFactory;
+                Task.Delay(5000).ContinueWith(_ => TryInitModel());
```
rayao: Sorry the Delay is for testing, to make the pipeline truly async. It's not required.

rayao:
I'm not quite sure if we should go this way. Another option is to for the model pipeline synchronous.

lewischeng-ms:
@rayao, based on our offline discussion, please change the PR accordingly and merge. Thanks！

Issue #323, opened 2016-03-16
Consider using a hook handler (IQueryExecutor) in publisher (WebAPI) to process the TotalCount


lewischeng-ms:
... as suggested by Mark.

Then this hook handler can call its inner handler which is typically the one from the provider.

rayao:
The problem is IQueryExecutor can't process ApiData based query, the input query must be processed by DefaultQueryHandler first.

rayao:
Oh do you mean add an IQueryExecutor implementation which call DefaultQueryHandler to translate the query, then call the inner provider IQueryExecutor?

lewischeng-ms:
@rayao, when RESTier calls the IQueryExecutor, the LINQ query has already been processed by the query pipeline (such as replacing ApiData, inspection, filtering, etc.) thus it should be ready to execute anyway. We just inject a query executor from Restier.WebApi to deal with "odata.count" (which is to send an additional count query) and then call the inner query executor. The good thing here is that we still save a lot of time from the count query because it does not need to go through the query pipeline again.

lewischeng-ms:
I know there could be some critical points we need to deal with but it could be a fairly good option which 1) moves the "odata.count" logic out of provider; 2) doesn't lower the performance much.

rayao:
Cool. Then perhaps we should ignore $top/skip/expand/count in WebApi, then in the WebApi QueryExecutor, generate 2 query base on the input query, 1 apply $top/skip/expand, the other 1 apply $count. Will that work?

rayao:
Let WebApi inject a service would be problematic under current design.

lewischeng-ms:
Yeah, there are a lot of benefit like these which we could possibly gain from this design. Again there are many things ahead that we need to work out. But it worths a try:-)

rayao:
Resolved by #329

PR #325, opened 2016-03-17
Remove ApiBuilder

rayao:
2 things need your attention:
- DefaultQueryExecutor made public, it's the only really default service, and that fact only benefits tests.
- `ApiConfiguration.Config<T>` added to register services from outside the API type. Optionally. Thus for example WebApi could register provider services.


src/Microsoft.Restier.Core/ApiConfiguration.cs
```diff
@@ -60,6 +66,23 @@ public IServiceProvider ServiceProvider
...
 
         /// <summary>
+        /// Adds a configuration procedure for API type <typeparamref name="T"/>.
+        /// </summary>
+        /// <typeparam name="T">The API type.</typeparam>
+        /// <param name="configurationCallback">
+        /// An action that will be called during the configuration of <typeparamref name="T"/>.
+        /// </param>
+        [CLSCompliant(false)]
+        public static void Configure<T>(Action<IServiceCollection> configurationCallback)
```
lewischeng-ms: Seems this method has no upstream caller and it is not overridable. Is it some kind of new design to facilitate the users?

rayao: I called it in ApiBase, ApiConfiguration creation logic, to allow service registration from outside, thus enable provider registered services.
Since it must be public it could also be used by users. If you think inappropriate I'll remove it.

src/Microsoft.Restier.Core/Query/DefaultQueryExecutor.cs
```diff
@@ -12,7 +12,7 @@ namespace Microsoft.Restier.Core.Query
     /// <summary>
     /// Default implementation for <see cref="IQueryExecutor"/>
     /// </summary>
-    internal class DefaultQueryExecutor : IQueryExecutor
+    public class DefaultQueryExecutor : IQueryExecutor
```
lewischeng-ms: It's OK for this PR. We may figure out another way to satisfy the test need.

lewischeng-ms:
Looks good to me. Thanks.

src/Microsoft.Restier.Core/ApiConfiguration.cs
lewischeng-ms: So it could have benefit if we want to inject a query executor from Restier.WebApi?

lewischeng-ms: I see. So as long as we call `ApiConfiguration.Configure<SomeApi>(do sth)` in the startup code, the callback will automatically get called the first time an `ApiConfiguration` instance for `SomeApi` is created. That may be a good solution for #323. Let's keep it and see how it works:-) Thanks for that!

rayao: Yeah exactly! :)

rayao:
Merged, than you Lewis!

PR #326, opened 2016-03-18
Prototype: $count processed by WebApi implemented QueryExecutor.

rayao:
#323 
I think it's basically like this, don't know whether meets your thoughts.

### DIFF
--- a/src/Microsoft.Restier.WebApi/Query/QueryExecutorWebApi.cs
+++ b/src/Microsoft.Restier.WebApi/Query/QueryExecutorWebApi.cs
@@ -1,0 +1,45 @@
+﻿// Copyright (c) Microsoft Corporation.  All rights reserved.
+// Licensed under the MIT License.  See License.txt in the project root for license information.
+
+using System.Linq;
+using System.Linq.Expressions;
+using System.Threading;
+using System.Threading.Tasks;
+using System.Web.OData.Extensions;
+using Microsoft.Restier.Core.Query;
+
+namespace Microsoft.Restier.WebApi.Query
+{
+    internal class QueryExecutorWebApi : IQueryExecutor
+    {
+        public IQueryExecutor Inner { get; set; }
+
+        public async Task<QueryResult> ExecuteQueryAsync<TElement>(
+            QueryContext context,
+            IQueryable<TElement> query,
+            CancellationToken cancellationToken)
+        {
+            var queryContext = context.ApiContext.GetApiService<WebApiContext>();
+            if (queryContext.QueryIncludeTotalCount == true)
+            {
+                var countQuery = ExpressionHelpers.GetCountableQuery(query);
+                var expression = countQuery.Expression;
+                expression = ExpressionHelpers.Count(expression, countQuery.ElementType);
+                var result = await ExecuteSingleAsync<long>(context, countQuery, expression, cancellationToken);
+
+                queryContext.Request.ODataProperties().TotalCount = result.Results.Cast<long>().Single();
+            }
+
+            return await Inner.ExecuteQueryAsync<TElement>(context, query, cancellationToken);
+        }
+
+        public Task<QueryResult> ExecuteSingleAsync<TResult>(
+            QueryContext context,
+            IQueryable query,
+            Expression expression,
+            CancellationToken cancellationToken)
+        {
+            return Inner.ExecuteSingleAsync<TResult>(context, query, expression, cancellationToken);
+        }
+    }
+}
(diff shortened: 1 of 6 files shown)


lewischeng-ms:
@chinadragon0515 Please help review Ray's design proposal together. Thanks.

rayao:
@lewischeng-ms @chinadragon0515 No hurry, and it's a prototype just to share my understanding of your intention.
Enjoy your holiday Lewis, it'll be a wonderful trip I believe. :)

lewischeng-ms:
Thanks @rayao :-)

src/Microsoft.Restier.WebApi/RestierController.cs
```diff
@@ -76,15 +76,11 @@ public IApi Api
             }
 
             IQueryable queryable = this.GetQuery();
-            QueryRequest queryRequest = new QueryRequest(queryable, this.includeTotalCount)
+            QueryRequest queryRequest = new QueryRequest(queryable, false)
             {
                 ShouldReturnCount = this.shouldReturnCount
             };
             QueryResult queryResult = await Api.QueryAsync(queryRequest, cancellationToken);
```
chinadragon0515: There maybe a race condition here? When queryAsync is executed, the context QueryIncludeTotalCount is not set yet, and cause QueryExecutorWebApi does not call count query?

src/Microsoft.Restier.WebApi/Query/QueryExecutorWebApi.cs
```diff
@@ -0,0 +1,45 @@
...
+using System.Linq;
+using System.Linq.Expressions;
+using System.Threading;
+using System.Threading.Tasks;
+using System.Web.OData.Extensions;
+using Microsoft.Restier.Core.Query;
+
+namespace Microsoft.Restier.WebApi.Query
+{
+    internal class QueryExecutorWebApi : IQueryExecutor
```
chinadragon0515: rename to WebApiQueryExecutor?

src/Microsoft.Restier.WebApi/Query/QueryExecutorWebApi.cs
```diff
@@ -0,0 +1,45 @@
...
+            IQueryable<TElement> query,
+            CancellationToken cancellationToken)
+        {
+            var queryContext = context.ApiContext.GetApiService<WebApiContext>();
+            if (queryContext.QueryIncludeTotalCount == true)
+            {
+                var countQuery = ExpressionHelpers.GetCountableQuery(query);
+                var expression = countQuery.Expression;
+                expression = ExpressionHelpers.Count(expression, countQuery.ElementType);
+                var result = await ExecuteSingleAsync<long>(context, countQuery, expression, cancellationToken);
```
chinadragon0515: General comment: I checked the code, seems ExecuteSingleAsync key logic will be countQuery.provider.execute(expression), then question will be whether we need two parameters here (query and expression) or rename query to QueryProvider?

src/Microsoft.Restier.WebApi/Query/QueryExecutorWebApi.cs
```diff
@@ -0,0 +1,45 @@
...
+            QueryContext context,
+            IQueryable<TElement> query,
+            CancellationToken cancellationToken)
+        {
+            var queryContext = context.ApiContext.GetApiService<WebApiContext>();
+            if (queryContext.QueryIncludeTotalCount == true)
+            {
+                var countQuery = ExpressionHelpers.GetCountableQuery(query);
+                var expression = countQuery.Expression;
+                expression = ExpressionHelpers.Count(expression, countQuery.ElementType);
```
chinadragon0515: What's the major purpose of this call? seems we make it generic with specific type, and make it as call, the expression is already object type which is enough for count, also I have not tested whether countQuery.ElementType can get more specific type here besides object.

src/Microsoft.Restier.WebApi/WebApiContext.cs
```diff
@@ -0,0 +1,14 @@
...
+
+namespace Microsoft.Restier.WebApi
+{
+    internal class WebApiContext
+    {
+        public HttpRequestMessage Request { get; set; }
+
+        public bool? QueryIncludeTotalCount { get; set; }
+    }
+}
```
chinadragon0515: In general, I prefer we do not introduce a new class with only one additional properties, can we just use request.Properties to store one more attribute to indicate need total count returned?

rayao:
A complete PR sent as #329, close this.
Sorry @chinadragon0515 for wasting your time to review.

PR #329, opened 2016-03-21
Handle $count in WebApi, through an injected IQueryExecutor

rayao:
WebApi publisher implements an IQueryExecutor to run an additional total count query.

### DIFF
--- a/src/Microsoft.Restier.Core/Query/DefaultQueryHandler.cs
+++ b/src/Microsoft.Restier.Core/Query/DefaultQueryHandler.cs
@@ -67,13 +67,6 @@ namespace Microsoft.Restier.Core.Query
             if (elementType != null)
             {
                 var query = visitor.BaseQuery.Provider.CreateQuery(expression);
-                if (query is EnumerableQuery)
-                {
-                    // Use the default query executor to handle query created
-                    // from IEnumerable<T> by calling query.ToList().
-                    executor = DefaultQueryExecutor.Instance;
-                }
-
                 var method = typeof(IQueryExecutor)
                     .GetMethod("ExecuteQueryAsync")
                     .MakeGenericMethod(elementType);
--- a/src/Microsoft.Restier.EntityFramework/DbApi.cs
+++ b/src/Microsoft.Restier.EntityFramework/DbApi.cs
@@ -60,14 +60,13 @@ namespace Microsoft.Restier.EntityFramework
         [CLSCompliant(false)]
         protected override IServiceCollection ConfigureApi(IServiceCollection services)
         {
-            services = base.ConfigureApi(services)
+            return base.ConfigureApi(services)
                 .CutoffPrevious<IModelBuilder>(ModelProducer.Instance)
                 .CutoffPrevious<IModelMapper>(new ModelMapper(typeof(T)))
                 .CutoffPrevious<IQueryExpressionSourcer, QueryExpressionSourcer>()
-                .CutoffPrevious<IQueryExecutor>(QueryExecutor.Instance)
+                .ChainPrevious<IQueryExecutor, QueryExecutor>()
                 .CutoffPrevious<IChangeSetPreparer, ChangeSetPreparer>()
-                .CutoffPrevious<ISubmitExecutor>(SubmitExecutor.Instance);
-            services
+                .CutoffPrevious<ISubmitExecutor>(SubmitExecutor.Instance)
                 .AddScoped<T>(sp =>
                 {
                     var dbContext = this.CreateDbContext(sp);
@@ -79,7 +78,6 @@ namespace Microsoft.Restier.EntityFramework
                     return dbContext;
                 })
                 .AddScoped<DbContext>(sp => sp.GetService<T>());
-            return services;
         }
 
         /// <summary>
--- a/src/Microsoft.Restier.EntityFramework/Query/QueryExecutor.cs
+++ b/src/Microsoft.Restier.EntityFramework/Query/QueryExecutor.cs
@@ -25,22 +25,12 @@ namespace Microsoft.Restier.EntityFramework.Query
     /// </summary>
     internal class QueryExecutor : IQueryExecutor
     {
-        static QueryExecutor()
-        {
-            Instance = new QueryExecutor();
-        }
-
-        private QueryExecutor()
-        {
-        }
-
-        /// <summary>
-        /// Gets the single instance of this query executor.
-        /// </summary>
-        public static QueryExecutor Instance { get; private set; }
+        public IQueryExecutor Inner { get; set; }
 
         /// <summary>
         /// Asynchronously executes a query and produces a query result.
+        /// This class only executes queries against EF provider, it'll
+        /// delegate other queries to inner IQueryExecutor.
         /// </summary>
         /// <typeparam name="TElement">
         /// The type of the elements in the query.
@@ -63,16 +53,17 @@ namespace Microsoft.Restier.EntityFramework.Query
             IQueryable<TElement> query,
             CancellationToken cancellationToken)
         {
-            long? totalCount = null;
-            if (context.Request.IncludeTotalCount == true)
+#if EF7
+            if (query.Provider is IAsyncQueryProvider)
+#else
+            if (query.Provider is IDbAsyncQueryProvider)
+#endif
             {
-                var countQuery = ExpressionHelpers.GetCountableQuery(query);
-                totalCount = await countQuery.LongCountAsync(cancellationToken);
+                return new QueryResult(
+                    await query.ToArrayAsync(cancellationToken));
             }
 
-            return new QueryResult(
-                await query.ToArrayAsync(cancellationToken),
-                totalCount);
+            return await Inner.ExecuteQueryAsync(context, query, cancellationToken);
         }
 
         /// <summary>
@@ -109,9 +100,14 @@ namespace Microsoft.Restier.EntityFramework.Query
 #else
             var provider = query.Provider as IDbAsyncQueryProvider;
 #endif
-            var result = await provider.ExecuteAsync<TResult>(
-                expression, cancellationToken);
-            return new QueryResult(new TResult[] { result });
+            if (provider != null)
+            {
+                var result = await provider.ExecuteAsync<TResult>(
+                    expression, cancellationToken);
+                return new QueryResult(new TResult[] { result });
+            }
+
+            return await Inner.ExecuteSingleAsync<TResult>(context, query, expression, cancellationToken);
         }
     }
 }
--- a/src/Microsoft.Restier.WebApi/HttpConfigurationExtensions.cs
+++ b/src/Microsoft.Restier.WebApi/HttpConfigurationExtensions.cs
@@ -9,10 +9,13 @@ using System.Web.Http;
 using System.Web.OData.Extensions;
 using System.Web.OData.Routing;
 using System.Web.OData.Routing.Conventions;
+using Microsoft.Extensions.DependencyInjection;
 using Microsoft.OData.Core;
 using Microsoft.OData.Edm;
 using Microsoft.Restier.Core;
+using Microsoft.Restier.Core.Query;
 using Microsoft.Restier.WebApi.Batch;
+using Microsoft.Restier.WebApi.Query;
 using Microsoft.Restier.WebApi.Routing;
 
 namespace Microsoft.Restier.WebApi
@@ -44,6 +47,11 @@ namespace Microsoft.Restier.WebApi
         {
             Ensure.NotNull(apiFactory, "apiFactory");
 
+            ApiConfiguration.Configure<TApi>(services =>
+            {
+                services.AddScoped<ODataCountOption>()
+                    .ChainPrevious<IQueryExecutor, QueryTotalCount>();
+            });
             using (var api = apiFactory())
             {
                 var model = api.GetModelAsync().Result;
--- a/src/Microsoft.Restier.WebApi/Query/ODataCountOption.cs
+++ b/src/Microsoft.Restier.WebApi/Query/ODataCountOption.cs
@@ -1,0 +1,23 @@
+﻿// Copyright (c) Microsoft Corporation.  All rights reserved.
+// Licensed under the MIT License.  See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.Restier.WebApi.Query
+{
+    internal class ODataCountOption
+    {
+        /// <summary>
+        /// Gets or sets a value indicating whether the total
+        /// number of items should be retrieved when the
+        /// result has been filtered using paging operators.
+        /// </summary>
+        /// <remarks>
+        /// Setting this to <c>true</c> may have a performance impact as
+        /// the data provider may need to execute two independent queries.
+        /// </remarks>
+        public bool IncludeTotalCount { get; set; }
+
+        public Action<long> SetTotalCount { get; set; }
+    }
+}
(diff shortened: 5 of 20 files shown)


src/Microsoft.Restier.EntityFramework/Query/QueryExecutor.cs
```diff
@@ -63,16 +53,17 @@ private QueryExecutor()
...
-                var countQuery = ExpressionHelpers.GetCountableQuery(query);
-                totalCount = await countQuery.LongCountAsync(cancellationToken);
+                return new QueryResult(
+                    await query.ToArrayAsync(cancellationToken));
             }
 
-            return new QueryResult(
-                await query.ToArrayAsync(cancellationToken),
-                totalCount);
+            return await Inner.ExecuteQueryAsync(context, query, cancellationToken);
```
rayao: Now it's provider's responsibility to delegate query to inner executor if the query is not against the specific query provider.

src/Microsoft.Restier.WebApi/HttpConfigurationExtensions.cs
```diff
@@ -44,6 +47,11 @@ public static class HttpConfigurationExtensions
         {
             Ensure.NotNull(apiFactory, "apiFactory");
 
+            ApiConfiguration.Configure<TApi>(services =>
+            {
+                services.AddScoped<ODataCountOption>()
```
rayao: The additional total count request is transferred via a scoped service. It's also could be done by adding subclasses to QueryRequest/QueryResult and then some members, but I'd like to make it really cross-cutting, and so possibly be reused independently. If you don't like it, I'll change to use inheritance.

src/Microsoft.Restier.WebApi/Query/QueryTotalCount.cs
```diff
@@ -0,0 +1,44 @@
...
+        public async Task<QueryResult> ExecuteQueryAsync<TElement>(
+            QueryContext context,
+            IQueryable<TElement> query,
+            CancellationToken cancellationToken)
+        {
+            var countOption = context.ApiContext.GetApiService<ODataCountOption>();
+            if (countOption.IncludeTotalCount)
+            {
+                var countQuery = ExpressionHelpers.GetCountableQuery(query);
+                var expression = ExpressionHelpers.Count(countQuery.Expression, countQuery.ElementType);
```
rayao: I still think reverse engineering the expression is not a perfect solution, it's prone to break if some changes made to WebApi query options builder. But it seems the only solution to save 1 more query.

src/Microsoft.Restier.Core/Query/DefaultQueryHandler.cs
```diff
@@ -67,13 +67,6 @@ internal static class DefaultQueryHandler
             if (elementType != null)
             {
                 var query = visitor.BaseQuery.Provider.CreateQuery(expression);
-                if (query is EnumerableQuery)
-                {
-                    // Use the default query executor to handle query created
-                    // from IEnumerable<T> by calling query.ToList().
-                    executor = DefaultQueryExecutor.Instance;
```
rayao: We can't do this anymore, it bypasses publisher registered executor.

src/Microsoft.Restier.EntityFramework/DbApi.cs
```diff
@@ -60,14 +60,13 @@ protected T DbContext
         [CLSCompliant(false)]
         protected override IServiceCollection ConfigureApi(IServiceCollection services)
         {
-            services = base.ConfigureApi(services)
+            return base.ConfigureApi(services)
                 .CutoffPrevious<IModelBuilder>(ModelProducer.Instance)
                 .CutoffPrevious<IModelMapper>(new ModelMapper(typeof(T)))
                 .CutoffPrevious<IQueryExpressionSourcer, QueryExpressionSourcer>()
-                .CutoffPrevious<IQueryExecutor>(QueryExecutor.Instance)
+                .ChainPrevious<IQueryExecutor, QueryExecutor>()
```
chinadragon0515: Do you mean we may have more inner queryexecutor? I have impression this will be the most inner QueryExecutor? Or we want to have DefaultQueryExecutor as most inner query executor?

src/Microsoft.Restier.EntityFramework/Query/QueryExecutor.cs
chinadragon0515: QueryExecutor is part of provider, do we have case we need to call Inner.ExecuteQueryAsync in current EF provider?

src/Microsoft.Restier.WebApi/Query/ODataCountOption.cs
```diff
@@ -0,0 +1,23 @@
...
+        /// </summary>
+        /// <remarks>
+        /// Setting this to <c>true</c> may have a performance impact as
+        /// the data provider may need to execute two independent queries.
+        /// </remarks>
+        public bool IncludeTotalCount { get; set; }
+
+        public Action<long> SetTotalCount { get; set; }
+    }
+}
```
chinadragon0515: You will discard previous PR 326?
In general, I prefer we do not introduce a new class with only one additional properties, can we just use request.Properties to store one more attribute to indicate need total count returned?
Also ODataCountOption will confuse user with some WebAPi and OData lib class name.

src/Microsoft.Restier.WebApi/Query/QueryTotalCount.cs
chinadragon0515: I agree like WebApi changes how expand works, then we may need to change code again.

The other comment: 
What's the major purpose of this call? seems we make it generic with specific type, and make it as call, the expression is already object type which is enough for count, also I have not tested whether countQuery.ElementType can get more specific type here besides object.

src/Microsoft.Restier.WebApi/Query/QueryTotalCount.cs
```diff
@@ -0,0 +1,44 @@
...
+            return await Inner.ExecuteQueryAsync<TElement>(context, query, cancellationToken);
+        }
+
+        public Task<QueryResult> ExecuteSingleAsync<TResult>(
+            QueryContext context,
+            IQueryable query,
+            Expression expression,
+            CancellationToken cancellationToken)
+        {
+            return Inner.ExecuteSingleAsync<TResult>(context, query, expression, cancellationToken);
```
chinadragon0515: General comment: I checked the code, seems ExecuteSingleAsync key logic will be countQuery.provider.execute(expression), then question will be whether we need two parameters here (query and expression) or rename query to QueryProvider?

src/Microsoft.Restier.WebApi/Query/QueryTotalCount.cs
```diff
@@ -0,0 +1,44 @@
...
+using System.Linq;
+using System.Linq.Expressions;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Restier.Core.Query;
+
+namespace Microsoft.Restier.WebApi.Query
+{
+    internal class QueryTotalCount : IQueryExecutor
+    {
```
chinadragon0515: I was thinking a more general name, now it is only total count, and in future, we can add more OData specified logic here.

src/Microsoft.Restier.EntityFramework/Query/QueryExecutor.cs
rayao: Since DefaultQueryHandler now calls the outermost (WebApi in this case) IQueryExecutor, for EnumerableQuery, EF provider can't handle it correctly, so it must delegate it to inner (DefaultQueryExecutor).

src/Microsoft.Restier.WebApi/Query/QueryTotalCount.cs
rayao: This is just a copy from DefaultQueryHandler, it merely appends a Count() call to the countQuery, so that ExecuteSingleAsync can process it.
The ElementType property should be identical to the element type of the stripped query.

src/Microsoft.Restier.EntityFramework/DbApi.cs
rayao: Yeah we'd have to make DefaultQueryExecutor the innermost, since we removed the executor replacement in DefaultQueryHandler, now provider IQueryExecutor could receive an EnumerableQuery, which can't be handled by provider, and so provider should delegate it to inner executor (DefaultQueryExecutor).

src/Microsoft.Restier.WebApi/Query/ODataCountOption.cs
rayao: Yeah, #326 is just a prototype, and not synced to head. Sorry for wasting your review time.
You mean HttpRequestMessage.Properties?, Is there anyway Restier query pipeline can get the HttpRequestMessage?

src/Microsoft.Restier.WebApi/Query/QueryTotalCount.cs
rayao: Yeah I also think the interface method is designed weird, maybe a query itself is enough, since both the query provider and an expression is within the IQueryable.

src/Microsoft.Restier.WebApi/Query/QueryTotalCount.cs
rayao: My own opinion is that if this has nothing to do with OData/WebApi, it'll be much easier to reuse under some other circumstances. So I intentionally isolated any OData/WebApi specific logic here.
But I'd like to follow your design if you want it to be a WebApi dedicated implementation.

chinadragon0515: This is something specific to OData now, so I was thinking some new name like ODataQueryExecutor.

src/Microsoft.Restier.WebApi/Query/QueryTotalCount.cs
chinadragon0515: OK, thanks

src/Microsoft.Restier.WebApi/Query/ODataCountOption.cs
chinadragon0515: I did some check, see two possibilities,
1. We have some service like ApiContext,  we can have it hold request and use this request.
2. We can just use ApiContext setProperty on thism and get ApiContext.

You can make a call to see whether we need to change, I still new to these code and still try to get familiar to these code.

src/Microsoft.Restier.WebApi/Query/QueryTotalCount.cs
chinadragon0515: We can make changes later, thanks

src/Microsoft.Restier.EntityFramework/Query/QueryExecutor.cs
chinadragon0515: So if it is EnumerableQuery, what will be the query provider? Also can you help to add some comments here for future maintain of code?

rayao: DefaultQueryProvider handles EnumerableQuery.
Comments already added to DbApi QueryExecutor for this.

src/Microsoft.Restier.WebApi/Query/ODataCountOption.cs
rayao: Since ApiContext now is also a scoped service, so attaching something to it basically has the same effect as registering a scoped service (like this ODataCountOption class).
And actually I'm planning to start a discussion about removing PropertyBag. As we have DI now, I'd discourage using any generic dictionary as data store, we'd use dedicated services instead. Of course eventually you guys will make the decision, but for now let's stop adding it to ApiContext property bag.
So let me summarize the design options here, actually we just need a decision on the mindset.
- Option1: We add cross-cutting feature services, that each serves a specific feature (like getting total count), but not bound to a specific platform or publisher (like OData/WebApi). This way we may result in multiple services and can choose among them to register.
- Option2: We add services that bound to specific platform and implement all features together, results in less services that can only work on the specific platform or publisher.

Since now we only have 1 such requirement, these 2 options are kind of indistinguishable, but later the difference will be more and more prominent.
So you'd like to go option 2 right? If so I'll rename classes accordingly and assume they're bound to OData/WebApi publisher.

src/Microsoft.Restier.EntityFramework/Query/QueryExecutor.cs
chinadragon0515: I got this, thanks.

src/Microsoft.Restier.WebApi/Query/ODataCountOption.cs
chinadragon0515: From Restier view, I see request to expand both provider part (Like mongoDB) and publisher part (Like Swagger),, but I do expect we will have more request on the provider part to support more data source to be exposed as OData service, and much less on the publisher part.
Also I was thinking different publisher may have different logic, like for count, OData does not consider top/skip but consider filter/search, for other protocol, it may not consider top, but consider skip/filter/search.
 
Based on this assumption, I choose option 2.  And if we do not want to introduce more logic into ApiContext, then may we have  class to be ODataQueryOption or similar. 
You are expert on this code (I just know the code about 2 weeks :)), you can make the call.
I will discuss with Lewis when he is back.

src/Microsoft.Restier.EntityFramework/Query/QueryExecutor.cs
rayao: I'll try to add some more comments to IQueryExecutor, thank you.

src/Microsoft.Restier.WebApi/Query/ODataCountOption.cs
rayao: Ok, I'll change class names accordingly. Your assumption makes more sense, thank you.

rayao:
Pushed, thank you.
I'll leave this open in case @lewischeng-ms has any concern on the changes made.
- IQueryExecutor behavior change, mainly affects data providers.
- Passes OData specific query option as a scoped service.

src/Microsoft.Restier.WebApi/Query/QueryTotalCount.cs
lewischeng-ms: @rayao @chinadragon0515 Conventionally in `Restier.WebApi`, we name things with the `Restier` prefix instead of the `OData` prefix, such as `RestierController`, `RestierPayloadValueConverter`, etc. This is to differentiate from the names in OData Web API and ODataLib. Thus to keep consistent, please follow this convention (e.g., `RestierQueryExecutor` and `RestierQueryExecutorOptions`). Besides I have no other concerns. Thanks!

chinadragon0515: @lewischeng-ms I was thinking we may have publisher like OneAPi,OpenAPI and so on, then we  may have OneAPIQueryOptions, OpenAPIQueryOptions and so on, and this is the reason I suggest name OData as prefix, if we want to differentiate from low level ODL lib class names, I would suggest RestierODataQueryExecutor, then in future we may have RestierOneAPIQueryExecutor,RestierOpenApiQueryExecutor, comments?

lewischeng-ms: Yeah, I think your suggestion is good given that we may support multiple protocols in the future. But I'd suggest we keep consistency on the naming for now (at least before GA) unless we decide to add `OData` to all `RestierXXX` classes. When the time comes that we need to implement a new publisher for OneAPI, we may have more considerations. Some classes (e.g., serializers and deserializers) are not shared by OData and OneAPI then we can give them different names as you suggested. But for some shared classes (e.g., this query executor because OneAPI also has count option), we may give them the same names. Anyway we need to figure out a distinguishable and consistent set of names at that time.

lewischeng-ms:
@rayao Please close this issue. We should have no further concern about this PR any more. Thanks.

rayao:
Thank you!

PR #333, opened 2016-03-24
Remove PropertyBag

rayao:
This is only a proposal, if you do like a property bag to ease data accesses, feel **free to reject** it.
IMHO general purpose dictionary is not as necessary after we have DI enabled, instead we'd better add services for specific purposes. If on the other hand some consumer does want it, that could be easily done by registering a dictionary as scoped service, and that's on demand.
Please beaware the **behavior change** here. Previously asserted roles are bound to InvocationContext, now as a scoped service is bound to ApiContext. That means, if a service, say IQueryExpressionInspector, invoked when processing query A, starts a new query B (in its Inspect method), then query B will **inherit all asserted roles** from query A, since they share a same ApiContext (the scope). Before this change every query has its **own asserted role set**. Please let me know whether the new behavior is acceptable.


rayao:
Thank you @chinadragon0515 , the trick removed cla-required. :)

rayao:
This is not needed, close and discard it.

PR #339, opened 2016-03-29
Design proposal: A more flexible configuration system.

rayao:
This is to give the consuming part more flexibility on configuration.
- Make ApiContext the main entry point, since in provider code, we actually only need ApiContext. Now that ApiBase is registered as a service, its disposal could be properly handled by service container, we don't have to keep an instance of ApiBase. Of course ApiBase is still an optional entry point.
- Conventions are changed to be able to target Api types other than ApiBase.
- Adapt ApiBase and DbApi to the new configuration system.
- Adapt ApiConfigurationAttribute to the new configuration system.
- Add NorthwindApi2 to demonstrate the new configuration system.
- All UTs still pass, even if targeting Northwind tests at NorthwindApi2.


src/Microsoft.Restier.Core/Api.Interfaces.cs
```diff
@@ -0,0 +1,55 @@
...
+        void Configure(IServiceCollection services, Type apiType);
+    }
+
+    public interface IApiInitializer
+    {
+        void Initialize(ApiConfiguration configuration);
+    }
+
+    public interface IApiContextConfigurator
+    {
```
rayao: Public methods in ApiConfigurationAttribute are grouped into these 3 interfaces, to run some customized logic in various stage.

src/Microsoft.Restier.Core/Api.Interfaces.cs
```diff
@@ -0,0 +1,55 @@
...
+        void Cleanup(ApiContext context);
+    }
+
+    [CLSCompliant(false)]
+    public interface IApiContextFactory
+    {
+        ApiContext CreateWithin(IServiceScope scope);
+    }
+
+    public sealed class ApiTypeInfo
```
rayao: Since we could potentially have multiple api types, this service is introduced to discover related info.
I suppose later we could establish some caches around it, like property/method cache, instead of using reflection in every individual service.

src/Microsoft.Restier.Core/Api.cs
```diff
@@ -0,0 +1,30 @@
...
+// Licensed under the MIT License.  See License.txt in the project root for license information.
+
+using System;
+using System.Threading;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Microsoft.Restier.Core
+{
+    public sealed class Api<TKey> where TKey : class
+    {
```
rayao: This class could be eliminated entirely, and let ApiConfiguration manage configuration cache.
Or maybe we'd leave this and move ApiBuilder cache here?

src/Microsoft.Restier.Core/ApiBase.cs
```diff
@@ -84,27 +82,27 @@ protected ApiConfiguration Configuration
...
-                                .AddScoped(apiType, sp => sp.GetService<ApiHolder>().Api)
-                                .AddScoped(sp => sp.GetService<ApiHolder>().Api)
-                                .AddScoped(sp => sp.GetService<ApiHolder>().Api.Context);
-                        }
-
-                        var configuration = this.CreateApiConfiguration(services);
-                        ApiConfiguratorAttribute.ApplyConfiguration(apiType, configuration);
-                        return configuration;
+                        var builder = new ApiBuilder()
+                            .AddInnerMost(ApiConfiguration.Configure(apiType).Configuration)
```
rayao: Here's a hypothetical usage of ApiBuilder, we could use it to build a partial configuration, and then merge into other ApiBuilders. Not sure if it's useful enough.

src/Microsoft.Restier.Core/ApiBase.cs
```diff
@@ -84,27 +82,27 @@ protected ApiConfiguration Configuration
...
+                                {
+                                    services.AddScoped<ApiHolder>()
+                                        .AddScoped(apiType, sp => sp.GetService<ApiHolder>().Api)
+                                        .AddScoped(sp => sp.GetService<ApiHolder>().Api);
+                                }
+                            });
+                        var serviceCollection = new ServiceCollection();
+                        ApiBuilder.DefaultInnerMost(serviceCollection);
+                        builder.Configuration(serviceCollection);
+                        ApiBuilder.DefaultOuterMost(serviceCollection);
```
rayao: These lines equal to builder.Build(serviceCollection).

src/Microsoft.Restier.Core/ApiBuilder.cs
```diff
@@ -0,0 +1,81 @@
+﻿// Copyright (c) Microsoft Corporation.  All rights reserved.
+// Licensed under the MIT License.  See License.txt in the project root for license information.
+
+using System;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Microsoft.Restier.Core
+{
+    public sealed class ApiBuilder
```
rayao: Sorry I brought back ApiBuilder, but it's completely different now. It is to organize the configuration in a highly flexible way.
Now configuration is split into **inner **and **outer **parts, and consumer could add configuration to all the 4 ends.

src/Microsoft.Restier.Core/ApiBuilderExtensions.cs
```diff
@@ -0,0 +1,165 @@
...
+                }
+            };
+            return config;
+        }
+
+        public static ApiBuilder UseAttributes(this ApiBuilder obj, Type apiType)
+        {
+            Ensure.NotNull(obj, "obj");
+
+            return obj.AddInnerTail(AttributesConfiguration(apiType));
```
rayao: Attributes default to inner tail.

src/Microsoft.Restier.Core/ApiBuilderExtensions.cs
```diff
@@ -0,0 +1,165 @@
...
+                ConventionBasedOperationProvider.ApplyTo(services, apiType);
+                ConventionBasedEntitySetFilter.ApplyTo(services, apiType);
+            };
+        }
+
+        public static ApiBuilder UseConventions(this ApiBuilder obj, Type apiType)
+        {
+            Ensure.NotNull(obj, "obj");
+
+            return obj.AddOuterMost(ConventionsConfiguration(apiType));
```
rayao: Conventions default to outer most. Of course consumer could add other configuration call to outer most, after applying conventions, so that override conventions.

src/Microsoft.Restier.Core/ApiConfiguration.cs
```diff
@@ -35,10 +39,8 @@ namespace Microsoft.Restier.Core
     /// </remarks>
     public class ApiConfiguration : PropertyBag
     {
-        private static ConcurrentDictionary<Type, Action<IServiceCollection>> configurations =
-            new ConcurrentDictionary<Type, Action<IServiceCollection>>();
-
-        private static Action<IServiceCollection> emptyConfig = _ => { };
+        private static ConcurrentDictionary<Type, ApiBuilder> configurations =
```
rayao: ApiConfiguration maintains a global type to ApiBuilder map, but ApiBuilder could also be used standalone. And then build standalone ApiConfiguration.

src/Microsoft.Restier.Core/ApiConfigurationExtensions.cs
```diff
@@ -14,6 +16,17 @@ public static class ApiConfigurationExtensions
         private const string IgnoredPropertiesKey = "Microsoft.Restier.Core.IgnoredProperties";
 
         /// <summary>
+        /// Creates an <see cref="ApiContext"/> configured by current <see cref="ApiConfiguration"/>.
+        /// </summary>
+        /// <param name="obj">The <see cref="ApiConfiguration"/>.</param>
+        /// <returns>An <see cref="ApiContext"/>.</returns>
+        public static ApiContext CreateContext(this ApiConfiguration obj)
```
rayao: ApiContext is created from ApiConfiguration directly, no need for Api type.

src/Microsoft.Restier.Core/ApiBase.cs
```diff
@@ -52,16 +52,14 @@ public ApiContext Context
 
                 if (this.apiContext == null)
                 {
-                    this.apiContext = this.CreateApiContext(
```
rayao: I can't adapt CreateApiContext virtual method, so I removed it.

src/Microsoft.Restier.Core/Conventions/ConventionBasedApiModelBuilder.cs
```diff
@@ -95,7 +95,8 @@ private IQueryable GetEntitySetQuery(QueryExpressionContext context)
                 object target = null;
                 if (!entitySetProperty.GetMethod.IsStatic)
                 {
-                    target = context.QueryContext.GetApiService<ApiBase>();
+                    target = context.QueryContext.ApiContext.
```
rayao: Conventions should be (optionally) applied to each Api type separately, so a conventional service instance should still handle a specific Api type.

src/Microsoft.Restier.EntityFramework/ApiBuilderExtensions.cs
```diff
@@ -0,0 +1,42 @@
...
+using Microsoft.Restier.EntityFramework.Model;
+using Microsoft.Restier.EntityFramework.Query;
+using Microsoft.Restier.EntityFramework.Submit;
+
+namespace Microsoft.Restier.EntityFramework
+{
+    [CLSCompliant(false)]
+    public static class ApiBuilderExtensions
+    {
+        public static IServiceCollection UseDbContext<T>(this IServiceCollection obj)
```
rayao: Add multiple DbContext type is also possible, just need modifications to DB related services.
Of course we should first ask "do we need"?

src/Microsoft.Restier.Samples.Northwind/Models/NorthwindApi.cs
```diff
@@ -114,4 +115,114 @@ private class NorthwindModelExtender : IModelBuilder
...
+    [Grant(ApiPermissionType.All, On = "ExpensiveProducts")]
+    [Grant(ApiPermissionType.All, On = "Orders")]
+    [Grant(ApiPermissionType.All, On = "Employees")]
+    [Grant(ApiPermissionType.All, On = "Regions")]
+    [Grant(ApiPermissionType.Inspect, On = "Suppliers")]
+    [Grant(ApiPermissionType.Read, On = "Suppliers")]
+    [Grant(ApiPermissionType.All, On = "ResetDataSource")]
+    public class NorthWindApi2
+    {
+        public static void Configure()
```
rayao: No virtual method to override like in ApiBase, but a static method seems not good enough either.

src/Microsoft.Restier.Security/ApiPolicyActivator.cs
```diff
@@ -50,10 +50,10 @@ public Expression Expand(QueryExpressionContext context)
                 return CallInner(context);
             }
 
-            var target = context.QueryContext.GetApiService<ApiBase>();
-            var entitySetProperty = target.GetType().GetProperties(
+            var apiTypes = context.QueryContext.GetApiService<ApiTypeInfo>().ApiTypes;
```
rayao: This is only to show the usage of ApiTypeInfo, each instance of ApiPolicyActivator should still be applied to a specific Api type.

rayao:
I believe you noticed that public surface is actually bigger, that's awkward...
But I think for common scenarios the number of classes to know basically remains same.
And some newly added types provide new or enhanced capabilities.

rayao:
- Revert `ApiBuilder and Api<T>`
- Enable customizing ApiConfiguration at 4 anchors.
- Enable creating ApiContext without ApiBase, i.e. any type could be API types.
- Add methods and extensions to ease ApiConfiguration creation, especially for ApiBase-free scenario.
- Configuration interfaces introduced in last commit are kept, since they're important for ApiBase-free configuration, and are compatible to ApiBase configuration.
- Update projects to reference xunit 2.1.0, add xunit VS runner 2.1.0.

src/Microsoft.Restier.Samples.Northwind/App_Start/WebApiConfig.cs
```diff
@@ -34,6 +45,50 @@ public static void Register(HttpConfiguration config)
             await config.MapRestierRoute<NorthwindApi>(
                 "NorthwindApi", "api/Northwind",
                 new RestierBatchHandler(server));
+
+            var configuration = new Lazy<ApiConfiguration>(
+                BuildNorthwind2Configuration, LazyThreadSafetyMode.ExecutionAndPublication);
+            await config.MapRestierRoute<NorthwindApi2>(
+                "NorthwindApi2", "api/Northwind2",
+                () => configuration.Value.CreateContext(),
```
rayao: Configuration code need not to be bound to Api type, if ApiBase-free.

rayao:
Nobody likes it, discard and close.

lewischeng-ms:
@rayao Sorry to give you that illusion. We like it. But we would like to postpone the decision until we have reviewed the public APIs of the configuration system to understand what need to be improved and what are still missing. We are most likely to do that next week or a bit later (since we have a three-day offsite next week). And during this period of time (this month), breaking changes for ODataLib are the top priority for the team so we may not have enough time to focus on any big new design of RESTier. But I have opened a related issue to keep track of this PR. We will get back here for sure. Thanks for your great idea!

rayao:
@lewischeng-ms Oh, sorry. I saw no response and assumed the proposal is not useful to you. I still have a local repo on home PC, would you allow if I submit a new PR? Of course with latest changes merged first.

lewischeng-ms:
That would be of great help! Thanks!

PR #345, opened 2016-03-31
Utilize DI for payload converter

### LINKED ISSUE #335: Use DI for ODataPayloadValueConverter
Rather than have users to implement their own model builder to register their `ODataPayloadValueConverter`. Use DI is much more flexible.

chinadragon0515:
Fix #335 , user can register his own payload converter.

### DIFF
--- a/src/Microsoft.Restier.WebApi/HttpConfigurationExtensions.cs
+++ b/src/Microsoft.Restier.WebApi/HttpConfigurationExtensions.cs
@@ -47,6 +47,7 @@ namespace Microsoft.Restier.WebApi
         {
             Ensure.NotNull(apiFactory, "apiFactory");
 
+            // ApiBase.ConfigureApi is called before this method is called.
             ApiConfiguration.Configure<TApi>(services =>
             {
                 services.AddScoped<RestierQueryExecutorOptions>()
@@ -55,7 +56,7 @@ namespace Microsoft.Restier.WebApi
             using (var api = apiFactory())
             {
                 var model = api.GetModelAsync().Result;
-                model.EnsurePayloadValueConverter();
+
                 var conventions = CreateRestierRoutingConventions(config, model, apiFactory);
 
                 if (batchHandler != null && batchHandler.ApiFactory == null)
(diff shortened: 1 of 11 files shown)


src/Microsoft.Restier.WebApi/HttpConfigurationExtensions.cs
```diff
@@ -47,15 +47,22 @@ public static class HttpConfigurationExtensions
...
             {
-                services.AddScoped<RestierQueryExecutorOptions>()
+                services = services.AddScoped<RestierQueryExecutorOptions>()
                     .ChainPrevious<IQueryExecutor, RestierQueryExecutor>();
+
+                // This means there is no customized converted registered
+                if (!services.HasService<ODataPayloadValueConverter>())
+                {
+                    services.AddSingleton<ODataPayloadValueConverter, RestierPayloadValueConverter>();
+                }
```
rayao: - Could be replaced with TryAddScoped
- converted -> converter

src/Microsoft.Restier.WebApi/HttpConfigurationExtensions.cs
```diff
@@ -65,6 +72,11 @@ public static class HttpConfigurationExtensions
 
                 var route = config.MapODataServiceRoute(
                     routeName, routePrefix, model, new DefaultODataPathHandler(), conventions, batchHandler);
+
+                // Customized converter should be added in ConfigureApi as service
+                var converter = api.Context.GetApiService<ODataPayloadValueConverter>();
```
rayao: If the converter will always be retrieved from IEdmModel later, we can simply add
`if (converter == null) converter = new RestierPalyloadValueConverter();`
and omit the default registration.

src/Microsoft.Restier.WebApi/HttpConfigurationExtensions.cs
lewischeng-ms: TryAddScoped...Awesome!

src/Microsoft.Restier.WebApi/HttpConfigurationExtensions.cs
lewischeng-ms: +1

lewischeng-ms:
LGTM

src/Microsoft.Restier.WebApi/HttpConfigurationExtensions.cs
chinadragon0515: Will omit this registration.

src/Microsoft.Restier.WebApi/HttpConfigurationExtensions.cs
chinadragon0515: Yeah, this also help to resolve an issue of some our test cases which calls Api configuration before calling map route, and cause null reference exception (no exception in normal case which calls map route first)

PR #348, opened 2016-04-04
Several PropertyBag-related refactors

### LINKED ISSUE #347: Use DI for PropertyBag

lewischeng-ms:
Now all parts of `PropertyBag` logic are organized as extension methods and internal classes into the DI container. The API of `ApiConfiguration`, `ApiContext` and `InvocationContext` are much cleaner and decoupled from the `PropertyBag` logic.

Also changed the `GetApiService<T>` series of methods to extension as well. But this is just a proposal.

This PR resolves #347


src/Microsoft.Restier.Core/ApiBase.cs
```diff
@@ -85,7 +85,10 @@ protected ApiConfiguration Configuration
                     apiType =>
                     {
                         IServiceCollection services = new ServiceCollection()
-                            .CutoffPrevious<IQueryExecutor>(DefaultQueryExecutor.Instance);
+                            .CutoffPrevious<IQueryExecutor>(DefaultQueryExecutor.Instance)
+                            .AddSingleton<ApiConfigurationExtensions.PropertyBag>()
+                            .AddScoped<ApiContextExtensions.PropertyBag>()
+                            .AddScoped<InvocationContextExtensions.PropertyBag>();
```
rayao: 1 scoped PropertyBag service would be enough since you already make InvocationContext a dedicated scope.

src/Microsoft.Restier.Core/ApiContextExtensions.cs
```diff
@@ -77,7 +201,9 @@ public static class ApiContextExtensions
 
             try
             {
-                var model = await builder.GetModelAsync(new InvocationContext(context), cancellationToken);
+                var buildContext = new InvocationContext(context);
+                var model = await builder.GetModelAsync(buildContext, cancellationToken);
+                buildContext.DisposeScope();
```
rayao: Maybe make InvocationContext disposable and then `using(...)` here.

src/Microsoft.Restier.Core/InvocationContextExtensions.cs
```diff
@@ -0,0 +1,185 @@
...
+        /// </summary>
+        /// <param name="context">
+        /// An invocation context.
+        /// </param>
+        /// <typeparam name="T">The service type.</typeparam>
+        /// <returns>The service instance.</returns>
+        public static T GetApiContextService<T>(this InvocationContext context) where T : class
+        {
+            Ensure.NotNull(context, "context");
+            return context.ApiContext.GetApiService<T>();
```
rayao: I'm inclined not to get services from ApiContext, if we decide to give InvocationContext its own scope, we'd better assume every service resolving is targeting InvocationContext, instead of ApiContext.
We only have to assign ApiHolder.Api for the scope of InvocationContext. Or if it's possible to be more aggressive, we could remove the ApiContext scope, and make InvocationContext the primary parameter of all those service calls. If we want more compatibility, maybe inherit InvocationContext from ApiContext?

src/Microsoft.Restier.Core/Query/DefaultQueryHandler.cs
```diff
@@ -58,7 +58,7 @@ internal static class DefaultQueryHandler
 
             // execute query
             QueryResult result;
-            var executor = context.GetApiService<IQueryExecutor>();
+            var executor = context.GetApiContextService<IQueryExecutor>();
```
rayao: Again I'm inclined to retrieve services from the InvocationContext scope, scopes are equally created, ApiContext scope isn't special, when compared to InvocationContext scope.

chinadragon0515: @rayao Could we save instance creation/dispose here? For example, as ApiContext may have multiple InnovacationContext, then in case of invocationContext.getService, it may get an object created and dispose several  times, but if we use invocationContext,parentContext.getService, only one object of one type is needed to be created?

rayao: Most of Restier services have singleton lifetime, so only very few objects will get created and disposed for a new scope.

src/Microsoft.Restier.Core/ApiBase.cs
lewischeng-ms: Exactly. But then I found it hard to determine the best place to put the "scoped" `PropertyBag`. An alternative would be to put the two `PropertyBag`s (one for `ApiConfiguration` and one for scope) into `PropertyBagBase.cs`. But to keep consistent, I bind each one to its own extension class because the `PropertyBag` logic are only offered through the extension classes. Do you have any suggestion?

src/Microsoft.Restier.Core/ApiContextExtensions.cs
lewischeng-ms: +1. But then I guess we also need to make `ApiContext` disposable to keep consistent? Anyway we can do this in a separate task.

src/Microsoft.Restier.Core/InvocationContextExtensions.cs
lewischeng-ms: Agree. I tried to get all services from `InvocationContext` however I met some unknown failures. I guess we need to change them to use `InvocationContext` **one by one** to see which services failed and why.

I think though now `ApiContext` scope looks unnecessary, `InvocationContext` and `ApiContext` serve for different purposes (per-service-call vs per-request). We can eliminate the `GetApiContextService<T>` here however we may still allow users to store some request-based services in the DI container and access them by calling `invocationContext.ApiContext.GetApiService<T>`.

src/Microsoft.Restier.Core/Query/DefaultQueryHandler.cs
lewischeng-ms: I will test each API service according to the discussion above.

rayao: The most problematic ones in my mind would be DbContext and ApiBase.
DbContext is scoped so it'll be created with every InvocationContext, I remember in submit pipeline there's a nested invocation to query pipeline, so they'll use different DbContext.
ApiBase is scoped so it'll be disposed with every InvocationContext. That means we'd better figure out a way to create an ApiBase for each InvocationContext.

src/Microsoft.Restier.Core/ApiBase.cs
rayao: Perhaps I didn't get you exactly. IMHO we just declare a internal PropertyBag class and add it as scoped service, then ApiConfig, ApiContext, InvocationContext just get it from their own service provider.

lewischeng-ms: Got it! Does it mean that even if we declared `PropertyBag` as scoped then in `ApiConfiguration` we can get one from the root scope which is actually the same as singleton?

rayao: Yeah exactly, root provider is also in a scope, although without ServiceScope object. Actually we should even dispose it if under some circumstances, we're done with an ApiConfiguration. The way is `((IDisposable)serviceProvider).Dispose()`

src/Microsoft.Restier.Core/Query/DefaultQueryHandler.cs
lewischeng-ms: Thanks for the reminder. I also meant only those API services **which were previously called hook handlers**. Cannot replace every `GetApiContextService<T>` with `GetApiService<T>`:-)

src/Microsoft.Restier.Core/ApiBase.cs
lewischeng-ms: Thank you! Opened an issue to track that work item #349.

rayao:
My suggestion is if by any means we could eliminate ApiContext scope, we'd better go that way. Giving both ApiContext and InvocationContext a dedicated scope is not a good idea, my gut feeling.

lewischeng-ms:
Yeah I have the same feeling now... We are still choosing specific scopes (per-request or per-invocation) to create a service in our code but this should also be specified (as part of the policy) by users. But if we have to remove one scope, I still prefer to preserve the `ApiContext` scope. Removing the `InvocationContext` scope currently will only affect `PropertyBag` and `RoleBasedAuthorizer` (which stores asserted roles per-invocation). But I think we can always find workarounds for them if there is a real customer ask? But I guess we can hardly implement per-request services (at least very costly) with only `InvocationContext` scope (for which there is already a real customer need). I am not saying that per-invocation don't make sense but per-request is more common in my opinion. Anyway, thanks for pointing that out. We do need a change.

rayao:
Actually I prefer ApiContext scope too, if only 1 must live.
I've even considered introducing an InvocationStack service, to store per-invocation data, and push/pop on invocation begin/ends.

lewischeng-ms:
OK. I will remove the `InvocationContext` scope for now and let's figure out the design for `PropertyBag` and `RoleBasedAuthorizer` later. Thanks.

src/Microsoft.Restier.Security/InvocationContextExtensions.cs
```diff
@@ -33,11 +33,11 @@ public static class InvocationContextExtensions
         {
             Ensure.NotNull(context, "context");
             Ensure.NotNull(role, "role");
-            var assertedRoles = context.GetProperty<List<string>>(AssertedRoles);
+            var assertedRoles = context.ApiContext.GetProperty<List<string>>(AssertedRoles);
```
lewischeng-ms: Temporary workaround: leverage the `PropertyBag` from the `ApiContext` scope to store the `AssertedRoles`. The consequence is the information is now per-request not per-invocation.

lewischeng-ms:
@rayao @chinadragon0515 @VikingsFan Thank you all for your help!

Issue #330, opened 2016-03-22
200 with empty collection should be returned for null collection on server side


chinadragon0515:
Now if a property is a collection, and collection is null but not empty collection, 500 will be returned. And this should return empty collection.

Also note for EF, it only supports navigation collection, it does not support collection of primitive/enum/complex.

chinadragon0515:
Null collection of primitive/enum/complex is fixed (note EF does not support collection of these now).
Null collection of nav is not fixed as EF query executor does not allow to compare collection of nav prop .

chinadragon0515:
Use a workaround and it is merged, close the issue

PR #351, opened 2016-04-07
Make collection of navigation could be null for in-memory data source

chinadragon0515:
Make when nav collection is null, In memory data source will return 200. part fix of issue #330

### DIFF
--- a/src/Microsoft.Restier.Core/Conventions/ConventionBasedEntitySetFilter.cs
+++ b/src/Microsoft.Restier.Core/Conventions/ConventionBasedEntitySetFilter.cs
@@ -22,6 +22,8 @@ namespace Microsoft.Restier.Core.Conventions
             this.targetType = targetType;
         }
 
+        public IQueryExpressionFilter Inner { get; set; }
+
         /// <inheritdoc/>
         public static void ApplyTo(
             IServiceCollection services,
@@ -29,16 +31,24 @@ namespace Microsoft.Restier.Core.Conventions
         {
             Ensure.NotNull(services, "services");
             Ensure.NotNull(targetType, "targetType");
-            services.CutoffPrevious<IQueryExpressionFilter>(new ConventionBasedEntitySetFilter(targetType));
+            services.ChainPrevious<IQueryExpressionFilter>(next => new ConventionBasedEntitySetFilter(targetType)
+            {
+                Inner = next,
+            });
         }
 
         /// <inheritdoc/>
         public Expression Filter(QueryExpressionContext context)
         {
             Ensure.NotNull(context, "context");
-            if (context.ModelReference == null)
+
+            if (Inner != null)
             {
-                return null;
+                var innerFilteredExpression = Inner.Filter(context);
+                if (innerFilteredExpression != null && innerFilteredExpression != context.VisitedNode)
+                {
+                    return innerFilteredExpression;
+                }
             }
 
             var dataSourceStubReference = context.ModelReference as DataSourceStubReference;
--- a/src/Microsoft.Restier.EntityFramework/Query/QueryExpressionFilter.cs
+++ b/src/Microsoft.Restier.EntityFramework/Query/QueryExpressionFilter.cs
@@ -1,0 +1,41 @@
+﻿// Copyright (c) Microsoft Corporation.  All rights reserved.
+// Licensed under the MIT License.  See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Linq.Expressions;
+using Microsoft.Restier.Core.Query;
+
+namespace Microsoft.Restier.EntityFramework.Query
+{
+    /// <summary>
+    /// A query expression filter to handle EF related logic.
+    /// </summary>
+    internal class QueryExpressionFilter : IQueryExpressionFilter
+    {
+        /// <inheritdoc/>
+        public Expression Filter(QueryExpressionContext context)
+        {
+            Ensure.NotNull(context, "context");
+
+            // TODO GitHubIssue#330: EF QueryExecutor will throw exception if check whether collections is null added.
+            // Error message likes "Cannot compare elements of type 'ICollection`1[[EntityType]]'.
+            // Only primitive types, enumeration types and entity types are supported."
+            if (context.VisitedNode.NodeType == ExpressionType.NotEqual)
+            {
+                var binaryExp = (BinaryExpression)context.VisitedNode;
+                var left = binaryExp.Left as MemberExpression;
+                var right = binaryExp.Right as ConstantExpression;
+                bool leftCheck = left != null && left.Type.IsGenericType
+                    && left.Type.GetGenericTypeDefinition() == typeof(ICollection<>);
+                bool rightCheck = right != null && right.Value == null;
+                if (leftCheck && rightCheck)
+                {
+                    return Expression.Constant(true);
+                }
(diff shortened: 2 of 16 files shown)


src/Microsoft.Restier.EntityFramework/Query/QueryExpressionFilter.cs
```diff
@@ -0,0 +1,41 @@
...
+            {
+                var binaryExp = (BinaryExpression)context.VisitedNode;
+                var left = binaryExp.Left as MemberExpression;
+                var right = binaryExp.Right as ConstantExpression;
+                bool leftCheck = left != null && left.Type.IsGenericType
+                    && left.Type.GetGenericTypeDefinition() == typeof(ICollection<>);
+                bool rightCheck = right != null && right.Value == null;
+                if (leftCheck && rightCheck)
+                {
+                    return Expression.Constant(true);
```
rayao: This is another provider specific requirement, better register the service in WebApi.

src/Microsoft.Restier.Core/Conventions/ConventionBasedEntitySetFilter.cs
```diff
@@ -22,23 +22,33 @@ private ConventionBasedEntitySetFilter(Type targetType)
...
+
         /// <inheritdoc/>
         public static void ApplyTo(
             IServiceCollection services,
             Type targetType)
         {
             Ensure.NotNull(services, "services");
             Ensure.NotNull(targetType, "targetType");
-            services.CutoffPrevious<IQueryExpressionFilter>(new ConventionBasedEntitySetFilter(targetType));
+            services.ChainPrevious<IQueryExpressionFilter>(next => new ConventionBasedEntitySetFilter(targetType)
```
lewischeng-ms: Just a proposal: we can access the target type inside `ConventionBaseEntitySetFilter` thus no need to pass in `targetType` and use this overload of `ChainPrevious<TService>`.

src/Microsoft.Restier.Core/Conventions/ConventionBasedEntitySetFilter.cs
```diff
@@ -22,23 +22,33 @@ private ConventionBasedEntitySetFilter(Type targetType)
...
         public Expression Filter(QueryExpressionContext context)
         {
             Ensure.NotNull(context, "context");
-            if (context.ModelReference == null)
+
+            if (Inner != null)
             {
-                return null;
+                var innerFilteredExpression = Inner.Filter(context);
+                if (innerFilteredExpression != null)
```
lewischeng-ms: Should be `innerFilteredExpression != context.VisitedNode` because later in `DefaultQueryHandler:
https://github.com/OData/RESTier/blob/master/src/Microsoft.Restier.Core/Query/DefaultQueryHandler.cs#L269-L281

If the returned node is different from the original one, the new one will be re-visited. But if the node is not changed, we can just go on with the original convention.

src/Microsoft.Restier.EntityFramework/Query/QueryExpressionFilter.cs
lewischeng-ms: This is an EF-specific issue. But WebApi does not necessarily know he is working with an EF provider. Thus in my opinion, it's OK to put it into `DbApi`.

test/ODataEndToEndTests/Microsoft.Restier.WebApi.Test.Services.TrippinInMemory/Models/Person.cs
```diff
@@ -17,6 +17,8 @@ public class Person
         [AutoExpand]
         public virtual ICollection<Person> Friends { get; set; }
 
+        public virtual ICollection<Trip> Trips { get; set; }
```
lewischeng-ms: Consider add some data for `Person.Trips` to test both null or not-null cases.

src/Microsoft.Restier.EntityFramework/Query/QueryExpressionFilter.cs
rayao: Sure. My concern is that it can only handle a specific pattern of expression building, and the pattern is WebApi specific.

lewischeng-ms: I see. But the overall process for RESTier to handle a query is that:
 - The publisher parses a request then build an initial LINQ
 - The provider is involved to process the LINQ and create a `QueryResult`
 - The publisher serializes the `QueryResult` to payload
IMO, LINQ is a common structured representation of the requests. No matter how a publisher builds it, it should always be fully interpretable to the `ExpressionVisitor` and also as a result to the API services in the providers (like `IQueryExpressionSourcer`, `IQueryExpressionFilter`, etc).

rayao: Ok got it. I was just worried that maybe the expression could be an Equal or operand could be in another order, etc.
Anyway please go ahead, it's just my own thought.

lewischeng-ms: I think your concern does make much sense. Well in theory, our EF provider should be able to handle all kinds of unsupported expressions in the LINQ but that's just a big work... Maybe we need to do that with a good design later but for now being able to solve a few specific issues is fine enough I think. Thanks anyway:-)

src/Microsoft.Restier.Core/Conventions/ConventionBasedEntitySetFilter.cs
chinadragon0515: I see, I will make it as next TODO item and make changes to other convention based filter.

src/Microsoft.Restier.Core/Conventions/ConventionBasedEntitySetFilter.cs
chinadragon0515: Now if inner filter has any logic, it will filter and return a not null expression, or it will return null just like what current filter does.
The logic here means if inner has some logic do the filter, then just return that one, or it will continue the processing with existing filter.

DO you want me to change inner filter logic make it will return original expression even it does nothing?

test/ODataEndToEndTests/Microsoft.Restier.WebApi.Test.Services.TrippinInMemory/Models/Person.cs
chinadragon0515: Sure, as not null case I have tested with Friends, so I did not add, I will some.

src/Microsoft.Restier.Core/Conventions/ConventionBasedEntitySetFilter.cs
lewischeng-ms: Typically a visitor returns the original node if no change happens, which allows us to give different meaning for a "null" value (like we want to abandon this node). The behavior of `Filter` here looks more like a mapping function thus IMO returning the original makes more sense.

test/ODataEndToEndTests/Microsoft.Restier.WebApi.Test.Services.TrippinInMemory/Models/Person.cs
lewischeng-ms: It's OK. I just want to make sure we cover every case.

PR #353, opened 2016-04-09
Add customizer to ApiConfiguration.

rayao:
- Enable customizing ApiConfiguration at 4 anchors.
- ApiBase will draw configurations from customizer when creating ApiConfiguration.
- Introduce a new overridable (ConfigureApiServices) to ApiBase, which gives descendants more freedom to choose from service sets.

### DIFF
--- a/src/Microsoft.Restier.Core/ApiBase.cs
+++ b/src/Microsoft.Restier.Core/ApiBase.cs
@@ -87,14 +87,8 @@ namespace Microsoft.Restier.Core
                         IServiceCollection services = new ServiceCollection()
                             .CutoffPrevious<IQueryExecutor>(DefaultQueryExecutor.Instance)
                             .AddScoped<PropertyBag>();
-                        services = this.ConfigureApi(services);
-                        ApiConfiguratorAttribute.ApplyApiServices(apiType, services);
 
-                        // Copy from pre-build registration.
-                        ApiConfiguration.Configuration(apiType)(services);
-
-                        // Make sure that all convention-based handlers are outermost.
-                        EnableConventions(services, apiType);
+                        services = this.ConfigureApiServices(apiType, services);
                         if (!services.HasService<ApiBase>())
                         {
                             services.AddScoped<ApiHolder>()
@@ -134,6 +128,31 @@ namespace Microsoft.Restier.Core
 
             this.Dispose(true);
             GC.SuppressFinalize(this);
+        }
+
+        /// <summary>
+        /// Configure the entire service set for this API. Descendants can override this method to choose
+        /// from the default service sets provided by RESTier.
+        /// </summary>
+        /// <param name="apiType">
+        /// CLR type of this <see cref="ApiBase"/>.
+        /// </param>
+        /// <param name="services">
+        /// The <see cref="IServiceCollection"/> with which to create an <see cref="ApiConfiguration"/>.
+        /// </param>
+        /// <returns>The <see cref="IServiceCollection"/>.</returns>
+        [CLSCompliant(false)]
+        protected virtual IServiceCollection ConfigureApiServices(Type apiType, IServiceCollection services)
+        {
+            var customizer = ApiConfiguration.Customize(apiType);
+            services.Apply(customizer.InnerMost);
+
+            return this.ConfigureApi(services)
+                .Apply(customizer.PrivateApi)
+                .UseAttributes(apiType)
+                .UseConventions(apiType)
+                .Apply(customizer.Overrides)
+                .Apply(customizer.OuterMost);
         }
 
         /// <summary>
@@ -195,36 +214,6 @@ namespace Microsoft.Restier.Core
                 this.apiContext.DisposeScope();
                 this.apiContext = null;
             }
-        }
-
-        /// <summary>
-        /// Enables code-based conventions for an API.
-        /// </summary>
-        /// <param name="services">
-        /// The <see cref="IServiceCollection"/> containing API service registrations.
-        /// </param>
-        /// <param name="targetType">
-        /// The type of a class on which code-based conventions are used.
-        /// </param>
-        /// <remarks>
-        /// This method adds hook points to the API configuration that
-        /// inspect a target type for a variety of code-based conventions
-        /// such as usage of specific attributes or members that follow
-        /// certain naming conventions.
-        /// </remarks>
-        private static void EnableConventions(
-            IServiceCollection services,
-            Type targetType)
-        {
-            Ensure.NotNull(services, "services");
-            Ensure.NotNull(targetType, "targetType");
-
-            ConventionBasedChangeSetAuthorizer.ApplyTo(services, targetType);
-            ConventionBasedChangeSetEntryFilter.ApplyTo(services, targetType);
-            services.CutoffPrevious<IChangeSetEntryValidator, ConventionBasedChangeSetEntryValidator>();
-            ConventionBasedApiModelBuilder.ApplyTo(services, targetType);
-            ConventionBasedOperationProvider.ApplyTo(services, targetType);
-            ConventionBasedEntitySetFilter.ApplyTo(services, targetType);
         }
 
         // Registered as a scoped service so that IApi and ApiContext could be exposed as scoped service.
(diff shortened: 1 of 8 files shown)


src/Microsoft.Restier.Core/ApiBase.cs
```diff
@@ -137,6 +131,31 @@ public void Dispose()
...
+            services.Apply(customizer.InnerMost);
+
+            return this.ConfigureApi(services)
+                .Apply(customizer.PrivateApi)
+                .UseAttributes(apiType)
+                .UseConventions(apiType)
+                .Apply(customizer.Overrides)
+                .Apply(customizer.OuterMost);
+        }
+
```
chinadragon0515: 1. I will prefer we keep ApiBase. It is straight forward and easy to extend and understand.
2. Prefer to get rid of apiType here. Maybe make services as property of the class.
3. Prefer only one interface to override (maybe remove configureApi)
4. For end user who overrides the Api, we allows several ways to register,
a).  All user to add more service at the head, and if user uses cutoff here, it means he override the service.
base. ConfigureApiServices()
service.ChainPrevious()
service.cutoff

b).  All user to add more service at the tail, and if user can only use cutoff (or other like addInstance here), but can not use ChainPrevious here.
service.ChainPrevious()
service.cutoff()
base. ConfigureApiServices()

c). Allow user do at both side
service. cutoff();
base.Configure()
service.cutoff()
service.chainprevious()

d). If user wants more control, this is what he will do in override configureAPiService
var customizer = ApiConfiguration.Customize(apiType);
customizer.Inner.append()
customizer.Out.prepend()
base.configureApi()

comments?

src/Microsoft.Restier.Core/ApiConfiguration.cs
```diff
@@ -64,32 +62,14 @@ public IServiceProvider ServiceProvider
...
         {
-            ApiConfiguration.configurations.AddOrUpdate(
-                typeof(T),
-                configurationCallback,
-                (type, existing) => existing + configurationCallback);
+            return Customize(typeof(TKey));
         }
 
-        internal static Action<IServiceCollection> Configuration(Type apiType)
+        public static Customizer Customize(Type keyType)
```
chinadragon0515: ApiServiceCustomer?

src/Microsoft.Restier.Core/ApiConfiguration.cs
```diff
@@ -120,5 +100,51 @@ internal TaskCompletionSource<IEdmModel> CompeteModelGeneration(out Task<IEdmMod
...
+
+        public sealed class Customizer
+        {
+            public Customizer()
+            {
+                InnerMost = new Anchor();
+                OuterMost = new Anchor();
+                Overrides = new Anchor();
+                PrivateApi = new Anchor();
+            }
```
chinadragon0515: Maybe remove Overrides
And we have 
InnerServiceChain (tail to service registered by RESTier code)
OutServiceChain (Head to the service registered by RESTier code)
InternalServiceChain (For RESTier code to use only).

src/Microsoft.Restier.EntityFramework/DbApi.cs
```diff
@@ -60,23 +60,18 @@ protected T DbContext
...
+                    var dbContext = instance.CreateDbContext(sp);
 #if EF7
                     // TODO GitHubIssue#58: Figure out the equivalent measurement to suppress proxy generation in EF7.
 #else
                     dbContext.Configuration.ProxyCreationEnabled = false;
 #endif
                     return dbContext;
                 })
-                .AddScoped<DbContext>(sp => sp.GetService<T>());
+                .UseDbContext<T>();
```
chinadragon0515: Maybe add to InternalServiceChain??

src/Microsoft.Restier.Core/ApiConfiguration.cs
```diff
@@ -120,5 +100,51 @@ internal TaskCompletionSource<IEdmModel> CompeteModelGeneration(out Task<IEdmMod
...
+
+            public Anchor OuterMost { get; private set; }
+
+            public Anchor PrivateApi { get; private set; }
+
+            public Anchor Overrides { get; private set; }
+        }
+
+        public sealed class Anchor
+        {
```
chinadragon0515: ApiServiceAnchor? ServiceAnchor?

src/Microsoft.Restier.Core/ApiBase.cs
rayao: 1. I'm not suggesting abandoning ApiBase, it's very easy and intuitive for a quick start. I'm just asking if we want it to be the only option, to host an API. And I'm suggesting another option, which might be useful to advanced scenarios.
2. services is for transient usage only, need not to be a member.
3. Yeah I also think that's a problem, but forcing descendants to add all those default service sets is also painful.
4. IMHO with customizer presents, the need for overriding ApiBase should be minimized, common consumer could just register services with customizer anchors.

src/Microsoft.Restier.Core/ApiConfiguration.cs
rayao: So user can't add services between provider and conventions?

src/Microsoft.Restier.Core/ApiBase.cs
chinadragon0515: 1.  Two notes
2. Ray's pull request......


thanks for your response, I think more about this and have a short discussion with Lewis,

The valid requests now,
1. Allow register a model builder before conversion model build to build entity set. And this could between Conversion model build and EF model builder to add more entity set.
2. Allow to register as a head model build so user can log a whole model.

Also behaviors are depended in some cases:
1. Inner most register may take no effect if same service we register later use cutoff.
2. Even user register as out most, it could be consider like inner most if user write code like 
inner.call()
customizedlogic();
3. Even in our code, when we call inner to do something, we may call it first or last.
Like model builder, we call inner first.
For QueryExecutor, with some check, and call inner later.
4. User can extends current service class, control whole logic. Then register out most without inner.

Any comments of these statements?

Based on these statements,
1. Inner most is not a concern as user can register out most and call inner first.

2. We must allow user to register out most, and I think we can achieve via 
base.Configure()
service.cutoff()
service.chainprevious()

3. We need to allow use to register after conversion (and before DBContext), user can do like 
service.AddCoreService() // Includes attributes and other registered services.
service.UseDbContext()// Optional
service.ChainPrevious()// User customized service
service.UserConversions()
service.ChainPrevious()// User customized service

I did not use other chains like inner/out/private, as I think it is hard for end user to understand what services are registered between each of these service chain.

4. For DbAPi, the key is query executor, its configuration will be like 
service.UseInternalService() // Includes attributes and other registered services.
serivice.UseDbContext()
service.UserConversions()

5. For RESTierQueryExecutor, a new extension method in WebApi project?

We can add more document of ConfigAPi and each user methods (maybe user change to AddEFService or something similar).
For internal service, 
And we allow user to have full control of the chain.

rayao: So eventually the customizer is not needed?

rayao: If we want simplification, IMHO existing configuration options (ConfigureApi) is enough, most users would just use the default configuration to expose a DbContext. The customizer is for advanced scenarios, and whoever uses it must have a knowledge of the default configuration and what each anchor means.
So IMHO we can first make the options fewer and then decide.

Option 1:
We make it extremely simple, only ConfigureApi and ConfigureApiServices (rename them as necessary). And thus we target 2 category of users.
- Most users who want only the basic scenario, no or very little customized configuration, which could be fulfilled by ConfigureApi.
- Some advanced users who use ConfigureApiServices to build completely customized configurations. ConfigureApiServices actually gives user full control on the chain.

Option 2:
We introduce some configuration APIs to ease those advanced users, so that in most cases they don't have to override configuration calls in ApiBase, they just use Customizer.

Option 3:
We go further to support advanced scenarios, allowing configurations created without ApiBase, that's #354

chinadragon0515: I create a pull request per discussion locally, can you help to review and let us know your comments? thanks in advance for your great help.

PR #354, opened 2016-04-09
A more flexible configuration system

rayao:
This is to give the consuming part more flexibility on configuration.
This change is **based on** #353, which introduces customizer and 4 anchors to configuration system. So #353 is relatively conservative and this one is more aggressive, but **full compatibility is guaranteed**. This is still addition to current system, no breaking to existing consumer.
Comparing to #353, this one adds features enabling fully customized configuration, and making ApiBase optional. The cost is wider public surface, to support new features.
So if we decide that ApiBase is mandatory, we could stop here and consider #353 instead.
- Enable creating ApiContext without ApiBase, i.e. any type could be API types.
- Add methods and extensions to ease ApiConfiguration creation, especially for ApiBase-free scenario.
- Introduce some configuration interfaces, which are important to ApiBase-free configuration, and are compatible to ApiBase configuration.
- Adapt ApiConfiguratorAttribute to the new configuration interfaces.
- Conventions are changed to be able to target Api types other than ApiBase.
- Make ApiContext the main entry point, since in provider code, we actually only need ApiContext. Now that ApiBase is registered as a service, its disposal could be properly handled by service container, we don't have to keep an instance of ApiBase. Of course ApiBase is still an optional entry point.
- Add NorthwindApi2 to demonstrate the new configuration system.
- All UTs still pass, even if targeting Northwind tests at NorthwindApi2.

### DIFF
--- a/src/Microsoft.Restier.Core/Api.Interfaces.cs
+++ b/src/Microsoft.Restier.Core/Api.Interfaces.cs
@@ -1,0 +1,34 @@
+﻿// Copyright (c) Microsoft Corporation.  All rights reserved.
+// Licensed under the MIT License.  See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Microsoft.Restier.Core
+{
+    [CLSCompliant(false)]
+    public interface IApiConfigurator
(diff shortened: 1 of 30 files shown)


rayao:
This is a re-submit of #339.
No need to hurry on this. I know you're busy on OData stuff.

PR #355, opened 2016-04-10
Update EF7 packages to RC1

rayao:
I'm too ignorant to find out that EF7 sqlserver package has been updated to RC1 months ago, with a new name. Sorry!


lewischeng-ms:
Looks good to me. Thanks for your contribution! But I'm curious about one thing... ASP.NET had already claimed to rename EF7 to EF Core 1.0... but looks like even the RC version is still 7.0...

rayao:
My understanding is that we'd have to expect more upcoming changes (or renames precisely). :(

PR #362, opened 2016-04-12
Truncate Northwind DB name if too long.

rayao:
EF7 Northwind sample can't work correctly with a single DB.
This change uses partial path (100 chars max) as prefix of database name, to avoid duplicate DB name.


lewischeng-ms:
I am OK with this change.

rayao:
Merged, thank you!

PR #368, opened 2016-04-19
Allow more customization for DI service

chinadragon0515:
For this PR, I mainly made such changes,
1. For the internal services, uses central class to register all services.
2. As core can not refer to WebApi, keep InternalServiceCallback for WebApi to add service, and then call added the services.
3. For end user, they can add as inner most or out most, and if then want to add service between some RESTier service, they just copy existing configureApi content and add service between RESTier service call. Northwind API is an example.
4. I keep conversion as we discussed. And make attributes as it is now(we will not ship this at first GA, it is not a concerns).

I have passed build and all test cases, still thinking some renames, any comments, let me know.

### DIFF
--- a/src/Microsoft.Restier.Core/ApiConfiguration.cs
+++ b/src/Microsoft.Restier.Core/ApiConfiguration.cs
@@ -34,7 +34,7 @@ namespace Microsoft.Restier.Core
     /// </remarks>
     public class ApiConfiguration
     {
-        private static ConcurrentDictionary<Type, Action<IServiceCollection>> _internalServicesCallback =
+        private static ConcurrentDictionary<Type, Action<IServiceCollection>> publisherServicesCallback =
             new ConcurrentDictionary<Type, Action<IServiceCollection>>();
 
         private static Action<IServiceCollection> emptyConfig = _ => { };
@@ -73,20 +73,20 @@ namespace Microsoft.Restier.Core
         /// An action that will be called during the configuration of <typeparamref name="TApi"/>.
         /// </param>
         [CLSCompliant(false)]
-        public static void AddInternalServices<TApi>(Action<IServiceCollection> configurationCallback)
+        public static void AddPublisherServices<TApi>(Action<IServiceCollection> configurationCallback)
              where TApi : ApiBase
         {
-            _internalServicesCallback.AddOrUpdate(
+            publisherServicesCallback.AddOrUpdate(
                 typeof(TApi),
                 configurationCallback,
                 (type, existing) => existing + configurationCallback);
         }
 
         [CLSCompliant(false)]
-        public static Action<IServiceCollection> GetInternalServiceCallback(Type apiType)
+        public static Action<IServiceCollection> GetPublisherServiceCallback(Type apiType)
         {
             Action<IServiceCollection> val;
-            if (_internalServicesCallback.TryGetValue(apiType, out val))
+            if (publisherServicesCallback.TryGetValue(apiType, out val))
             {
                 return val;
             }
--- a/src/Microsoft.Restier.Core/ApiConfiguration.cs
+++ b/src/Microsoft.Restier.Core/ApiConfiguration.cs
@@ -34,7 +34,7 @@ namespace Microsoft.Restier.Core
     /// </remarks>
     public class ApiConfiguration
     {
-        private static ConcurrentDictionary<Type, Action<IServiceCollection>> configurations =
+        private static ConcurrentDictionary<Type, Action<IServiceCollection>> _internalServicesCallback =
             new ConcurrentDictionary<Type, Action<IServiceCollection>>();
 
         private static Action<IServiceCollection> emptyConfig = _ => { };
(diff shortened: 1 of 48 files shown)


src/Shared/EdmHelpers.cs
```diff
@@ -5,7 +5,7 @@
 using System.Globalization;
 using System.Linq;
 using Microsoft.OData.Edm.Library;
-using Microsoft.Restier.Core.Properties;
+using Microsoft.Restier.WebApi.Properties;
```
chinadragon0515: This one I consider more changes, as it refer to a resource properties, and resource class is auto generated and it is internal, we can not really share the class, I was thinking just move it to WebApi project and no other class are using it now.
Comments?

src/Microsoft.Restier.Core/ApiConfiguration.cs
```diff
@@ -34,7 +34,7 @@ namespace Microsoft.Restier.Core
     /// </remarks>
     public class ApiConfiguration
     {
-        private static ConcurrentDictionary<Type, Action<IServiceCollection>> configurations =
+        private static ConcurrentDictionary<Type, Action<IServiceCollection>> _internalServicesCallback =
```
VikingsFan: remove "_" or use "_" in other places?

src/Microsoft.Restier.EntityFramework/DbApi.cs
```diff
@@ -59,39 +59,20 @@ protected T DbContext
...
-        /// The underlying DbContext used by this API.
-        /// </returns>
-        protected virtual T CreateDbContext(IServiceProvider serviceProvider)
-        {
-            return Activator.CreateInstance<T>();
+            // Add EF related services
+            services.AddDbContextServices<T>();
+
+            // This is used to add the publisher's services
+            // TODO, will think about a better way
```
VikingsFan: will we create issue for those TODO ?

src/Microsoft.Restier.WebApi/Routing/HttpConfigurationExtensions.cs
```diff
@@ -47,11 +43,10 @@ public static class HttpConfigurationExtensions
         {
             Ensure.NotNull(apiFactory, "apiFactory");
 
-            // ApiBase.ConfigureApi is called before this method is called.
-            ApiConfiguration.Configure<TApi>(services =>
+            // This will be added a service add callback which is added in ApiBase.ConfigureApi method.
```
VikingsFan: can't understand this sentence

src/Microsoft.Restier.EntityFramework/DbApi.cs
```diff
@@ -59,39 +59,20 @@ protected T DbContext
...
-        /// The service container of the currently being created <see cref="ApiContext"/>.
-        /// </param>
-        /// <returns>
-        /// The underlying DbContext used by this API.
-        /// </returns>
-        protected virtual T CreateDbContext(IServiceProvider serviceProvider)
-        {
-            return Activator.CreateInstance<T>();
+            // Add EF related services
+            services.AddDbContextServices<T>();
```
rayao: Is it intentional to add EF services after conventions? That's different from now.

src/Microsoft.Restier.WebApi/Model/ApiModelBuilder.cs
```diff
@@ -11,16 +11,17 @@
...
 
-namespace Microsoft.Restier.Core.Conventions
+namespace Microsoft.Restier.WebApi.Model
 {
     /// <summary>
     /// A convention-based API model builder that extends a model, maps between
     /// the model space and the object space, and expands a query expression.
     /// </summary>
-    internal class ConventionBasedApiModelBuilder
+    internal class ApiModelBuilder
```
rayao: This requires consumer to publish on WebApi to use this convention (it's still a convention, functionally), or at least add a reference to Restier.WebApi. It seems weird to me.

src/Microsoft.Restier.Core/ApiConfiguration.cs
chinadragon0515: dash is recommended by IDE, will keep as it is now.

VikingsFan: I still think we should keep consistence, if you want to keep it please update other places

src/Microsoft.Restier.EntityFramework/DbApi.cs
chinadragon0515: Yes, we had some discussion, we consider conversion should be core code, and core service is always been added first.
If some conversion depends on other services, we need to think whether the conversion is really in core.

This is the reason we move Conversion based model builder into WebApi project as we think it is OData specific.

src/Microsoft.Restier.WebApi/Model/ApiModelBuilder.cs
chinadragon0515: This is used to builder some model for OData, we have some discussion, we do not think other protocol can reuse this code as we do not think there are two protocol shares the same model. This is the reason we move to WebApi.

If user  wants to use OData with RESTier, then we expect user always have WebApi publisher, if user uses other protocol, we do not expect they will need this service.

src/Microsoft.Restier.EntityFramework/DbApi.cs
rayao: Does that mean any service touching EdmModel should be in WebApi?

src/Microsoft.Restier.WebApi/Model/ApiModelBuilder.cs
rayao: Does any other protocol could possibly reuse EdmModel?

src/Microsoft.Restier.EntityFramework/DbApi.cs
chinadragon0515: Depends on how we define EdmModel, if we think it is OData specific, then Yes.
But every protocol should have a model, we need an abstraction of model, we do not have this yet. So core code still have reference to EdmModel class which is part of OData protocol. 

In future, we expect we have some abstract level of model, and core has no reference of protocol specific code.

Your comments or suggestions?

src/Microsoft.Restier.WebApi/Model/ApiModelBuilder.cs
chinadragon0515: I can not say for sure now, but if some protocol reuse it, it will make two protocol very similar, like one possible is OneApi, we use same code based to support it and OData.

src/Microsoft.Restier.EntityFramework/DbApi.cs
chinadragon0515: One more comment is we have a plan to remove these conversion code when WebApi 6.0 is ready, we will use WebApi model builder directly without any code from RESTier.

PR #379, opened 2016-04-26
Prevent WebApi services from being registered for multiple times.

rayao:
In Northwind test AddWebApiServices could possibly be called more than once, causing several model extenders registered and results in wrong EDM model.
I'm curious why Northwind tests still pass, but Northwind.EF7 tests aren't that lucky.


chinadragon0515:
do we know how it get registered several times? Like how it get called to registered twice? Will user may get it registered twice too?

I am OK with fix but also prefer we try to avoid it been called twice at all.

rayao:
UTs run in a same domain and in parallel, in Northwind tests almost every case call `WebApiConfig.RegisterNorthwind(config, server)` which effectively register web api services multiple times. It won't affect a real production environment, it's the fault of the way we author tests.

chinadragon0515:
thanks, feel free to commit.

rayao:
Done

PR #381, opened 2016-04-27
Update public API per discussion

chinadragon0515:
Update some public API

### DIFF
--- a/src/Microsoft.Restier.WebApi/RestierController.cs
+++ b/src/Microsoft.Restier.WebApi/RestierController.cs
@@ -48,7 +48,7 @@ namespace Microsoft.Restier.WebApi
         /// <summary>
         /// Gets the API associated with this controller.
         /// </summary>
-        public ApiBase Api
+        internal ApiBase Api
         {
             get
             {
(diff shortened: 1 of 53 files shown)


src/Microsoft.Restier.WebApi/RestierController.cs
```diff
@@ -48,7 +48,7 @@ public class RestierController : ODataController
         /// <summary>
         /// Gets the API associated with this controller.
         /// </summary>
-        public ApiBase Api
+        internal ApiBase Api
```
lewischeng-ms: maybe private because users won't inherit `RestierController` or call `RestierController.Api`.

test/Microsoft.Restier.Core.Tests/Api.Tests.cs
```diff
@@ -48,7 +48,7 @@ private class TestModelMapper : IModelMapper
                 string namespaceName, string name,
                 out Type relevantType)
             {
-                relevantType = typeof(DateTime);
+                relevantType = typeof (DateTime);
```
lewischeng-ms: we don't want extra space between `typeof`/type casting and the left parenthesis. this is Resharper's default convention but not our coding style.

lewischeng-ms:
Looks good to me otherwise.

test/Microsoft.Restier.Core.Tests/Api.Tests.cs
chinadragon0515: Thanks, fixed

src/Microsoft.Restier.WebApi/RestierController.cs
chinadragon0515: Make it priviate

Issue #380, opened 2016-04-27
Int64 could not be used as key


cilerler:
*Even #123 says issue resolved, it is partially correct.  PUT still throws following error where PATCH, POST, DELETE, GET works fine after the fix*

### Assemblies affected
**0.5.0-beta.*  

### Actual result
```
{
  "error": {
    "code": "",
    "message": "An error has occurred.",
    "innererror": {
      "message": "Unsupported type for property: ID.",
      "type": "System.NotSupportedException",
[...]

PR #384, opened 2016-04-28
Fix issues when push entity whose key is long type

chinadragon0515:
### Issues
*This pull request fixes issue #380 .*  

### Description
Make put can work with key with type long(int64)

- [ x ] Test cases added

### DIFF
--- a/src/Microsoft.Restier.EntityFramework/Submit/ChangeSetPreparer.cs
+++ b/src/Microsoft.Restier.EntityFramework/Submit/ChangeSetPreparer.cs
@@ -230,6 +230,12 @@ namespace Microsoft.Restier.EntityFramework.Submit
                 return (TimeSpan)timeOfDayValue;
             }
 
+            // In case key is long type, when put an entity, key value will be from key parsing which is type of int
+            if (value is int && type == typeof(long))
+            {
+                return Convert.ToInt64(value, CultureInfo.InvariantCulture);
+            }
+
             return value;
         }
     }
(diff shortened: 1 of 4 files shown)


src/Microsoft.Restier.EntityFramework/Submit/ChangeSetPreparer.cs
```diff
@@ -230,6 +230,12 @@ private static object ConvertToEfValue(Type type, object value)
                 return (TimeSpan)timeOfDayValue;
             }
 
+            // In case key is long type, when put an entity, key value will be from key parsing which is type of int
+            if (value is int && type == typeof(long))
```
rayao: Could type be `Nullable<long>`?

chinadragon0515: No, as this is only used for key which is parsed by RESTier, and for other update properties, they are parsed by WebAPi, and type is always correct which I have verified.

test/ODataEndToEndTests/Microsoft.Restier.WebApi.Test.Services.Trippin/Models/Person.cs
```diff
@@ -26,7 +26,7 @@ public class Person
         [MaxLength(26), MinLength(1)]
         public string LastName { get; set; }
 
-        public int? Age { get; set; }
+        public long? Age { get; set; }
```
lewischeng-ms: `Age` is not key property?

lewischeng-ms: If it is to test the logic modified in `ChangeSetPreparer`, please make sure there IS some test data whose `Age` values are null.

lewischeng-ms:
Looks good to me for the other part.

test/ODataEndToEndTests/Microsoft.Restier.WebApi.Test.Services.Trippin/Models/Person.cs
chinadragon0515: Yes, I make some test specified null for Age now.

Issue #301, opened 2016-02-25
[Design] Use DI for the serializer provider and all the serializers


lewischeng-ms:
So that users can easily replace them with their customized ones. Now it is so hard to override the serializer behavior.

rayao:
I'm not very familiar with WebApi configuration flow, so maybe this question is stupid.
How the serializer/deserializer providers get context info? It seems to me RestierFormatting attribute applies to RestierController, so it only gets a controller descriptor, but although every Api uses RestierController type, different Api may have some specific serializer/deserializer, and that means the serializer providers must know route info (to get an Api instance for example). During attribute initialization there's only controller specific configuration settings, so it's impossible there. And at runtime, only GetODataPayloadSerializer gets routing info, GetEdmTypeSerializer gets nothing contextual, same for deserializer.

lewischeng-ms:
I think your concerns are reasonable. `GetEdmTypeSerializer` and `GetEdmTypeDeserializer` won't get the context info HOWEVER each single serializer/deserializer will get the context info from a parameter called `writeContext` when being called (e.g., https://github.com/OData/RESTier/blob/master/src/Microsoft.Restier.WebApi/Formatter/Serialization/RestierComplexTypeSerializer.cs). But what I am really worried about here is how to access the `ApiBuilder` and the `ApiConfiguration`.

rayao:
You're right, serializer/deserializer have access to RequestMessage, with ApiFactory in Properties.
It seems to me we'd have to move the Api into Request.Properties also.

ivanko2000:
"Pinging" this issue in case it went on the back burner...

I am currently trying to implement a deserializer for an entity with binary property (so when POSTing a new instance one could pass the JSON for entity definition and binary content for one of the properties in "multipart/form-data"). So far I haven't been successful in making it work, due to a combination of me being a newbie and the above mentioned
> Now it is so hard to override the serializer behavior

chinadragon0515:
@ivanko2000 thanks for your patience, Lewis is busy on other critical items, I discussed with him and take this from him today, it is my TODO items in the next several days. And plan to be part of 0.5.0 release.

chinadragon0515:
@lewischeng-ms @rayao I am investigating this item and thinking about the design we may have,

Facts,
1. Format is initialized when application is started, and a formatter need to be set.
2. For the serializer provider and serializer, there are some cross reference like for EntityTypeSerializer it needs Provider for further get other needed serializer.

[...]

PR #386, opened 2016-04-29
Allow serializer provider can be customized with DI services

chinadragon0515:
### Issues
*This pull request fixes issue #301.*  

### Description
Make serializer and deserializer provider as DI service, and only works for RESTier controller.
In case of error or user defined controller, default provider will be used.

- [ x ] Test cases added

### Additional work necessary
Document is needed.

### DIFF
--- a/src/Microsoft.Restier.WebApi/Formatter/Deserialization/DefaultRestierDeserializerProvider.cs
+++ b/src/Microsoft.Restier.WebApi/Formatter/Deserialization/DefaultRestierDeserializerProvider.cs
@@ -11,11 +11,23 @@ using Microsoft.Restier.WebApi.Results;
 namespace Microsoft.Restier.WebApi.Formatter.Deserialization
 {
     /// <summary>
-    /// The default serializer provider.
+    /// The default deserializer provider.
     /// </summary>
-    internal class DefaultRestierDeserializerProvider : DefaultODataDeserializerProvider
+    public class DefaultRestierDeserializerProvider : DefaultODataDeserializerProvider
     {
         private RestierEnumDeserializer enumDeserializer;
+        private static readonly DefaultRestierDeserializerProvider _instance = new DefaultRestierDeserializerProvider();
+
+        /// <summary>
+        /// Gets the default instance of the <see cref="DefaultRestierDeserializerProvider"/>.
+        /// </summary>
+        internal static DefaultRestierDeserializerProvider SingletonInstance
+        {
+            get
+            {
+                return _instance;
+            }
+        }
 
         /// <summary>
         /// Initializes a new instance of the <see cref="DefaultRestierDeserializerProvider" /> class.
--- a/src/Microsoft.Restier.WebApi/Formatter/Deserialization/RestierDeserializerProviderProxy.cs
+++ b/src/Microsoft.Restier.WebApi/Formatter/Deserialization/RestierDeserializerProviderProxy.cs
@@ -1,0 +1,67 @@
+﻿// Copyright (c) Microsoft Corporation.  All rights reserved.
+// Licensed under the MIT License.  See License.txt in the project root for license information.
+
+using System;
+using System.Net.Http;
+using System.Web.OData.Formatter.Deserialization;
+using System.Web.OData.Formatter.Serialization;
+using Microsoft.OData.Edm;
+using Microsoft.Restier.Core;
+using Microsoft.Restier.WebApi.Results;
+
+namespace Microsoft.Restier.WebApi.Formatter.Deserialization
+{
+    /// <summary>
+    /// The deserializer provider proxy which get real provider to implement the logic.
+    /// </summary>
+    internal class RestierDeserializerProviderProxy : ODataDeserializerProvider
+    {
+        private ApiBase api;
+
+        /// <summary>
+        /// Gets an <see cref="ODataDeserializer"/> for the given type.
+        /// The proxy provider will get the real provider to return the serializer.
+        /// </summary>
+        /// <param name="model">The EDM model.</param>
+        /// <param name="type">The CLR type.</param>
+        /// <param name="request">The request being deserialized.</param>
+        /// <returns>An <see cref="ODataDeserializer"/> that can deserialize the given type.</returns>
+        public override ODataDeserializer GetODataDeserializer(IEdmModel model, Type type, HttpRequestMessage request)
+        {
+            this.api = request.GetApiInstance();
+
+            if (this.api != null)
+            {
+                ODataDeserializerProvider provider = api.Context.GetApiService<ODataDeserializerProvider>();
+                if (provider != null)
+                {
+                    return provider.GetODataDeserializer(model, type, request);
+                }
+            }
+
+            // In case user uses his own controller or NonFound error for request
+            return DefaultRestierDeserializerProvider.SingletonInstance.GetODataDeserializer(model, type, request);
+        }
+
+        /// <summary>
+        /// Gets the <see cref="ODataEdmTypeDeserializer"/> for the given EDM type.
+        /// The proxy provider will get the real provider to return the serializer.
+        /// </summary>
+        /// <param name="edmType">The EDM type.</param>
+        /// <returns>An <see cref="ODataEdmTypeDeserializer"/> that can deserialize the given EDM type.</returns>
+        public override ODataEdmTypeDeserializer GetEdmTypeDeserializer(IEdmTypeReference edmType)
+        {
+            if (this.api != null)
+            {
+                ODataDeserializerProvider provider = api.Context.GetApiService<ODataDeserializerProvider>();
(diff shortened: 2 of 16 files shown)


src/Microsoft.Restier.WebApi/Formatter/Deserialization/DefaultRestierDeserializerProvider.cs
```diff
@@ -11,11 +11,23 @@
...
 {
     /// <summary>
-    /// The default serializer provider.
+    /// The default deserializer provider.
     /// </summary>
-    internal class DefaultRestierDeserializerProvider : DefaultODataDeserializerProvider
+    public class DefaultRestierDeserializerProvider : DefaultODataDeserializerProvider
     {
         private RestierEnumDeserializer enumDeserializer;
+        private static readonly DefaultRestierDeserializerProvider _instance = new DefaultRestierDeserializerProvider();
```
rayao: Is this class guaranteed to work in multi-threaded environment?

src/Microsoft.Restier.WebApi/RestierController.cs
```diff
@@ -55,6 +55,7 @@ public ApiBase Api
                 if (this.api == null)
                 {
                     this.api = this.Request.GetApiFactory().Invoke();
+                    this.Request.SetApiInstance(this.api);
```
rayao: I suppose this happens after request message deserialization?

src/Microsoft.Restier.WebApi/HttpRequestMessageExtensions.cs
```diff
@@ -17,6 +17,7 @@ internal static class HttpRequestMessageExtensions
     {
         private const string ChangeSetKey = "Microsoft.Restier.Submit.ChangeSet";
         private const string ApiFactoryKey = "Microsoft.Restier.Core.ApiFactory";
+        private const string ApiInstanceKey = "Microsoft.Restier.Core.ApiInstance";
```
lewischeng-ms: test github codeflow:-)

src/Microsoft.Restier.WebApi/Formatter/Deserialization/DefaultRestierDeserializerProvider.cs
chinadragon0515: Make change to use Lazy

src/Microsoft.Restier.WebApi/RestierController.cs
chinadragon0515: Yeah, I move the setting into controller select and did some test with post, it is called before deserializer.

src/Microsoft.Restier.WebApi/Formatter/Deserialization/RestierDeserializerProviderProxy.cs
```diff
@@ -0,0 +1,61 @@
...
+        /// <param name="request">The request being deserialized.</param>
+        /// <returns>An <see cref="ODataDeserializer"/> that can deserialize the given type.</returns>
+        public override ODataDeserializer GetODataDeserializer(IEdmModel model, Type type, HttpRequestMessage request)
+        {
+            this.api = request.GetApiInstance();
+
+            if (this.api != null)
+            {
+                ODataDeserializerProvider provider = api.Context.GetApiService<ODataDeserializerProvider>();
+                return provider.GetODataDeserializer(model, type, request);
```
lewischeng-ms: check null before calling.

src/Microsoft.Restier.WebApi/Formatter/Deserialization/DefaultRestierDeserializerProvider.cs
```diff
@@ -11,11 +11,24 @@
...
 {
     /// <summary>
-    /// The default serializer provider.
+    /// The default deserializer provider.
     /// </summary>
-    internal class DefaultRestierDeserializerProvider : DefaultODataDeserializerProvider
+    public class DefaultRestierDeserializerProvider : DefaultODataDeserializerProvider
     {
         private RestierEnumDeserializer enumDeserializer;
+        private static readonly Lazy<DefaultRestierDeserializerProvider> lazy
```
lewischeng-ms: not necessary to use `Lazy`. Static instance will be created and initialized the first time it's being used.

src/Microsoft.Restier.WebApi/Formatter/Serialization/RestierSerializerProviderProxy.cs
```diff
@@ -0,0 +1,63 @@
...
+            IEdmModel model,
+            Type type,
+            HttpRequestMessage request)
+        {
+            this.api  = request.GetApiInstance();
+
+            if (this.api != null)
+            {
+                ODataSerializerProvider provider = api.Context.GetApiService<ODataSerializerProvider>();
+                return provider.GetODataPayloadSerializer(model, type, request);
```
lewischeng-ms: also here in this file.

src/Microsoft.Restier.WebApi/Routing/RestierRoutingConvention.cs
```diff
@@ -57,7 +57,8 @@ public string SelectController(ODataPath odataPath, HttpRequestMessage request)
                 return null;
             }
 
-            request.SetApiFactory(apiFactory);
+            // Create ApiBase instance
+            request.SetApiInstance(apiFactory.Invoke());
```
lewischeng-ms: This pattern looks much better than before. THanks!

lewischeng-ms:
Looks good to me otherwise.

src/Microsoft.Restier.WebApi/Formatter/Deserialization/DefaultRestierDeserializerProvider.cs
rayao: I mean if DefaultRestierDeserializerProvider is thread safe? This static instance could be called at the same time by multiple request?

src/Microsoft.Restier.WebApi/ServiceCollectionExtensions.cs
```diff
@@ -25,6 +29,10 @@ public static IServiceCollection AddWebApiServices<T>(this IServiceCollection se
             services.AddSingleton<ODataQuerySettings>(querySettingFactory);
             services.AddSingleton<ODataValidationSettings>();
 
+            // Make serializer and deserializer provider as DI services
+            services.AddSingleton<ODataSerializerProvider, DefaultRestierSerializerProvider>();
+            services.AddSingleton<ODataDeserializerProvider, DefaultRestierDeserializerProvider>();
```
rayao: Should be TryAddSingleton in case consumer add provider before applying this.

src/Microsoft.Restier.WebApi/Formatter/Deserialization/DefaultRestierDeserializerProvider.cs
lewischeng-ms: I think you are right. Let's use `LazyThreadSafetyMode.ExecutionAndPublication` for `Lazy` then.

rayao: As you said Lewis, static instance has synchronization guarantee for its construction, so we don't need a publication protection, what I mean is if class DefaultRestierDeserializerProvider itself is thread safe? All its members including inherited ones? I've just checked, it seems the provider was used as singleton (for a controller) before this change, so now I think it is thread safe, and a static member is enough I believe.

lewischeng-ms: Got it. Thanks for the tip:-)

src/Microsoft.Restier.WebApi/ServiceCollectionExtensions.cs
chinadragon0515: Updated,

src/Microsoft.Restier.WebApi/Formatter/Deserialization/RestierDeserializerProviderProxy.cs
chinadragon0515: Add the check.

src/Microsoft.Restier.WebApi/Formatter/Serialization/RestierSerializerProviderProxy.cs
chinadragon0515: Add the check

src/Microsoft.Restier.WebApi/Formatter/Deserialization/DefaultRestierDeserializerProvider.cs
chinadragon0515: thanks for clarification.

PR #400, opened 2016-05-12
Removed the Default Value of the property

cilerler:
### Issues
There is no active issue yet.

### Description
**build.cmd** 's current logic picks **MSBuild12** first which doesn't support Default Value on Properties

- [  ] Test cases added
- [  ] Build and test with one-click build and test script passed

### DIFF
--- a/src/Microsoft.Restier.WebApi/Model/OperationAttribute.cs
+++ b/src/Microsoft.Restier.WebApi/Model/OperationAttribute.cs
@@ -36,6 +36,6 @@ namespace Microsoft.Restier.WebApi.Model
         /// If a operation does not have side effect, it means it is a function.
         /// If a operation has side effect, it means it is a action.
         /// </summary>
-        public bool HasSideEffects { get; set; } = false;
+        public bool HasSideEffects { get; set; }
     }
 }


chinadragon0515:
@cilerler we define the default value to make sure in the model, the Operation attribute is default for a function but not action.
And as our readme indicate, our code is develop and build with VS2015. 

So I do not plan to merge this request now.

Can I know what cause you to build by yourself ? Can you use VS2015 to build?  If now, I will consider a default constructor and set default value in constructor.

cilerler:
@chinadragon0515 since there is no NuGet packages with **0.5.0beta** around yet,I had to generate internal packages to place it into a project.  Looks like I had **MSBUILD12** in my system so **build.cmd** picked-up that one and throw an error on the line with default value.  If you won't remove it, then we need to update the **build.cmd** and remove the [part](https://github.com/OData/RESTier/blob/master/build.cmd#L10-L14)

chinadragon0515:
@cilerler after some discussion, we found we have a final signed build server is still VS2013, and c# always have false default value for bool type. I just remove the default value after property, very appreciate your help on this.

PR #401, opened 2016-05-12
Update wrong project filename

cilerler:
### Description
Project file name changed from WebApi to OData.  This file had the old project filename.

- [  ] Test cases added


chinadragon0515:
@cilerler thanks for your contribution, I just merge the fix.

PR #404, opened 2016-05-18
Integrate EFCore rc2

rayao:
### Issues
EFCore RC2 has just released.

### Description
Integrate EFCore rc2 libs.
**Breaking change**: ServiceCollection.AddInstance is replaced by SC.AddSingleton in rc2.

- [   ] Test cases added
- [   ] Build and test with one-click build and test script passed.
There're 2 UT cases fail consistently on my box, with or without this PR.

### Additional work necessary
None.


src/Microsoft.Restier.Core/packages.config
```diff
@@ -1,6 +1,18 @@
...
+  <package id="System.ComponentModel" version="4.0.1-rc2-24027" targetFramework="net45" />
+  <package id="System.Diagnostics.Debug" version="4.0.11-rc2-24027" targetFramework="net45" />
+  <package id="System.Globalization" version="4.0.11-rc2-24027" targetFramework="net45" />
+  <package id="System.Linq" version="4.1.0-rc2-24027" targetFramework="net45" />
+  <package id="System.Linq.Expressions" version="4.0.11-rc2-24027" targetFramework="net45" />
+  <package id="System.Reflection" version="4.1.0-rc2-24027" targetFramework="net45" />
+  <package id="System.Resources.ResourceManager" version="4.0.1-rc2-24027" targetFramework="net45" />
+  <package id="System.Runtime.Extensions" version="4.1.0-rc2-24027" targetFramework="net45" />
+  <package id="System.Threading" version="4.0.11-rc2-24027" targetFramework="net45" />
+  <package id="System.Threading.Tasks" version="4.0.11-rc2-24027" targetFramework="net45" />
```
chinadragon0515: What are these packages for? If there are not related DI or EF7, I prefer not include RC version now.

src/Microsoft.Restier.EntityFramework/Model/ModelProducer.cs
```diff
@@ -4,7 +4,11 @@
 using System;
 using System.Collections.Generic;
 using System.Collections.ObjectModel;
+#if EF7
+using Microsoft.EntityFrameworkCore;
+#else
 using System.Data.Entity;
+#endif
```
chinadragon0515: We add such code in EF project class, and remove EF7 related class, the question I have is can we do for all EF7 project class and make us does not EF7 project at all? Or it is a big effort?

chinadragon0515:
Ray, I add two comments, and can you help to merge pull request first as I am in middle of change namespace and project name which almost change all the class, I want to merge your changes to continue.

src/Microsoft.Restier.Core/packages.config
rayao: No worry about this. On a full .NET framework (on Windows), these dependencies result in System and System.Core. You can confirm that by examining the references of the project.

src/Microsoft.Restier.EntityFramework/Model/ModelProducer.cs
rayao: I'm not quite sure what you mean, but EFCore does have some API changes which incompatible to EF6, so those projects exist, if we want to keep only 1 copy of Restier code, perhaps we must add some wrapper layer for EF.

rayao:
Rolling build failed, on NuGet restore stage. It seems NuGet client version must be upgraded for this change.

Issue #117, opened 2015-07-17
[Protocol] Support Entity Type inheritance


karataliu:

lewischeng-ms:
Also verify type inheritance in URL works after this feature is completed.

chinadragon0515:
The issues will fix in several steps,
0. Model build derived type (PR #402 
1. Query with derived type (PR #410 )
2. Query options work with derived type
3. Patch with derived type
4. Post with derived type
5. Delete with derived type

And note derived type is only for entity type, EF does not support complex type inheritance now, and it is not supported by RESTier now.

chinadragon0515:
All supports is added, close the issue.

PR #412, opened 2016-05-27
Derivedtype support part 3

chinadragon0515:
Issues

This pull request fixes part of issue #117

Description

Add support of derived type patch/update/delete

### DIFF
--- a/src/Microsoft.Restier.Core/Query/QueryExpressionContext.cs
+++ b/src/Microsoft.Restier.Core/Query/QueryExpressionContext.cs
@@ -122,7 +122,7 @@ namespace Microsoft.Restier.Core.Query
             var resultType = method.ReturnType.FindGenericType(typeof(IEnumerable<>));
             if (sourceType == resultType)
             {
-                return new QueryModelReference(source.EntitySet,source.Type);
+                return new QueryModelReference(source.EntitySet, source.Type);
             }
 
             // source is a sequence of T1 and output is a sequence of T2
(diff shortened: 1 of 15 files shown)


src/Microsoft.Restier.Core/Query/QueryExpressionContext.cs
```diff
@@ -199,29 +199,33 @@ private QueryModelReference ComputeModelReference()
...
                 {
                     methodCall = node as MethodCallExpression;
-                    if (methodCall != null)
+                    if (methodCall == null)
                     {
-                        modelReference = this.GetModelReferenceForNode(node);
-                        if (modelReference != null)
+                        continue;
+                    }
+                    modelReference = this.GetModelReferenceForNode(node);
```
lewischeng-ms: blank line after closing bracket

lewischeng-ms: also the lines below (211, 220)

src/Microsoft.Restier.Core/Submit/ChangeSetItem.cs
```diff
@@ -148,16 +151,18 @@ public class DataModificationItem : ChangeSetItem
         /// </param>
         public DataModificationItem(
             string entitySetName,
-            string entityTypeName,
+            Type entityType,
```
lewischeng-ms: how about `expectedEntityType` (the expected entity type from the model)  and `actualEntityType`?

src/Microsoft.Restier.Providers.EntityFramework/Submit/ChangeSetInitializer.cs
```diff
@@ -40,6 +40,13 @@ internal class ChangeSetInitializer : IChangeSetInitializer
             {
                 object strongTypedDbSet = dbContext.GetType().GetProperty(entry.EntitySetName).GetValue(dbContext);
                 Type entityType = strongTypedDbSet.GetType().GetGenericArguments()[0];
+
+                // This means request entity is sub type of entity type
+                if (entry.ActualEntityType != null && entityType != entry.ActualEntityType)
```
lewischeng-ms: could `ActualEntityType` be null? if not, we should add the enforcement somewhere.

src/Microsoft.Restier.Publishers.OData/RestierController.cs
```diff
@@ -108,9 +108,19 @@ private ApiBase Api
                 throw new NotImplementedException(Resources.InsertOnlySupportedOnEntitySet);
             }
 
+            // In case of type inherience, the actual type will be different from entity type
```
lewischeng-ms: spell

src/Microsoft.Restier.Publishers.OData/RestierController.cs
```diff
@@ -108,9 +108,19 @@ private ApiBase Api
...
+            var entityType = path.EdmType;
+            var actualEntityType = path.EdmType;
+            if (edmEntityObject.ActualEdmType != null)
+            {
+                entityType = edmEntityObject.ExpectedEdmType;
```
lewischeng-ms: also "expected" here

lewischeng-ms:
LGTM otherwise

src/Microsoft.Restier.Core/Query/QueryExpressionContext.cs
chinadragon0515: Updated

src/Microsoft.Restier.Core/Submit/ChangeSetItem.cs
chinadragon0515: That's better, updated.

src/Microsoft.Restier.Providers.EntityFramework/Submit/ChangeSetInitializer.cs
chinadragon0515: In delete case, it will be null now. We do not set the value for delete case.
I add this also not breaking any existing logic like test cases which does not have type inheritance and does not set this.
Let me know if you still prefer some enforcement.

src/Microsoft.Restier.Publishers.OData/RestierController.cs
chinadragon0515: fixed, thanks.

src/Microsoft.Restier.Publishers.OData/RestierController.cs
chinadragon0515: updated.

Issue #415, opened 2016-05-30
Make DefaultODataPathHandler as DI service so user can replace and support slash


chinadragon0515:
refer to issue #414

chinadragon0515:
merged, and close the issue.

Issue #421, opened 2016-06-06
Support key value with alias


chinadragon0515:
When request an entity with key, support key value as alias, and also validate back slash and slash support.

This is depending on https://github.com/OData/odata.net/issues/570

And a temp solution is added with issue https://github.com/OData/RESTier/issues/414

robertmclaws:
Thanks for your input! We are resetting the project as we move to (finally!) ship 1.0. While we are closing this for now, we will re-address legacy feature requests at a later date. Please be sure to check out the latest nightly builds (and the upcoming 1.0 Beta 2 build) and let us know if you run into any issues. Thanks again!

Issue #414, opened 2016-05-27
Escape for slashes in path


danroot:
With a path such as this:  /service/Users('Some%5Cuser') a 404 is always returned.  What is the best way to fix this or override the issue within Restier?

### Assemblies affected
RESTier 0.4.0-rc2

### Reproduce steps
Build a service on a EF context with a table with string key
Ensure a record with key 'Some\user'
browse to /service/Users('Some%5Cuser')

### Expected result
Expect a return value.

### Actual result
404

### Additional details
This is due to a "by design" behavior of Web API OData.  An explanation and a workaround is provided in this sample:
https://github.com/OData/ODataSamples/tree/master/WebApi/v4/ODataPathAndSlashEscapeSample 

However, I'm unclear how to implement this in Restier.

chinadragon0515:
@danroot we have not exposed IPahHandler as DI service in RESTier yet, the workaround is not simple as we do in Web Api now.

To workaround this one,
1. Create a class similar to https://github.com/OData/ODataSamples/blob/master/WebApi/v4/ODataPathAndSlashEscapeSample/ProductODataService/Extensions/PathAndSlashEscapeODataPathHandler.cs

2. In your WebApiConfig.cs file, copy methods from https://github.com/OData/RESTier/blob/master/src/Microsoft.Restier.Publishers.OData/Routing/HttpConfigurationExtensions.cs), and replace "new DefaultODataPathHandler()" with "new CustomizedODataPathHandler()".

Let me know if you have more questions, in the meanwhile, I create #415 to make this replacement same as other service, and will be delivered in next release.

chinadragon0515:
This sample does not work, stay tuned.
I am working on some workable solution now, both temp solution and fix in code.

chinadragon0515:
Per comments from web api team, IIS will escape back slash and this is not configurable.

A temp solution with end to end test case is added with #415, refer to it for detail on how to enable , but back slash and slash is not recommended used in the key as Web API suggests, as there may become a security hole.

Also for permanent solution, it is tracked with #421 , key value will be as query options and refer to URL via key alias.

I close this issue now, any questions or comments, let us know.

PR #422, opened 2016-06-06
Temp solution for URI with backslash and slash

chinadragon0515:
### Issues
*This pull request fixes issue #415 and it is a temp solution for #414 

### Description
Make URL can parse slash or backslash in key of entity

- [ x ] Test cases added

### Additional work necessary
This is a temp solution, need to revisit when key as alias is supported by ODL, and this is tracked with https://github.com/OData/RESTier/issues/421

### DIFF
--- a/src/Microsoft.Restier.Publishers.OData/Routing/HttpConfigurationExtensions.cs
+++ b/src/Microsoft.Restier.Publishers.OData/Routing/HttpConfigurationExtensions.cs
@@ -60,8 +60,17 @@ namespace Microsoft.Restier.Publishers.OData.Routing
                     batchHandler.ApiFactory = apiFactory;
                 }
 
+
+                // Customized path handler should be added in ConfigureApi as service
+                // Allow to handle URL encoded slash (%2F), and backslash(%5C) with customized handler
+                var handler = api.Context.GetApiService<IODataPathHandler>();
+                if (handler == null)
+                {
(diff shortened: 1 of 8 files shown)


src/Microsoft.Restier.Publishers.OData/Routing/HttpConfigurationExtensions.cs
```diff
@@ -60,8 +60,17 @@ public static class HttpConfigurationExtensions
                     batchHandler.ApiFactory = apiFactory;
                 }
 
+
+                // Customized path handler should be added in ConfigureApi as service
+                // Allow to handle URL encoded slash (%2F), and backslash(%5C) with customized handler
+                var handler = api.Context.GetApiService<IODataPathHandler>();
+                if (handler == null)
+                {
+                    handler = new DefaultODataPathHandler();
```
lewischeng-ms: It's better to register the default type `DefaultODataPathHandler` to the container.

lewischeng-ms:
LGTM otherwise.

src/Microsoft.Restier.Publishers.OData/Routing/HttpConfigurationExtensions.cs
chinadragon0515: Per discussion, we keep as it is now.

Issue #378, opened 2016-04-26
Add operation support with RESTier own controller


chinadragon0515:
Current in RESTier, the operation (function and action) are supported with additional controller and WebApi ODataRoute attribute, add support with RESTier controller.

Need to support both bounded and unbounded operation, and also support in batch request.

Before this is added, consumer can create separate controller to support operation, docs will be added for this part.

chinadragon0515:
The current way to support operation, refer to document http://odata.github.io/RESTier/#03-03-Operation, thanks.

chinadragon0515:
This issue will be resolve in several steps,
1. unbound function with zero or one parameter.
2.unbound function with more than one parameters.
3.unbound action with zero or one parameter and without request body
4.unbound action with zero or one parameter and with request body
5.unbound action with more than one parameters and without request body
6.unbound action with more than one parameters and without request body
7. bound function support
8. bound action support

chinadragon0515:
1,2,7 are supported with PR #424, and note path segment after function or query options for function is not supported yet.

chinadragon0515:
Action part are supported with PR https://github.com/OData/RESTier/pull/453, all code has merged.
Keep this issue open for document purpose only.

chinadragon0515:
document updated.

PR #424, opened 2016-06-06
Add function support (part 1)

chinadragon0515:
### Issues
*This pull request fixes issue #378

### Description
User does not need additional controller for operation route.

- [ x ] Test cases added

### Additional work necessary
Will need to update document


src/Microsoft.Restier.Publishers.OData/Formatter/Deserialization/DeserializationHelpers.cs
```diff
@@ -0,0 +1,95 @@
...
+namespace Microsoft.Restier.Publishers.OData.Formatter.Deserialization
+{
+    /// <summary>
+    /// Get clr type from payload.
+    /// </summary>
+    internal static class DeserializationHelpers
+    {
+        internal static object ConvertValue(object oDataValue, Type expectedReturnType, ref IEdmTypeReference propertyType, ApiBase api)
+        {
+            var model = api.GetModelAsync().Result;
```
rayao: This is dangerous, for a truly async model pipeline this could hang forever.

chinadragon0515: I change these method to async and use await, this should avoid deadlock, let me know if you have other suggestions on this part. I am new to c# :)

src/Microsoft.Restier.Publishers.OData/RestierController.cs
```diff
@@ -525,6 +550,127 @@ private ODataPath GetPath()
...
+
+        private async Task<IQueryable> ExecuteBoundFunction(BoundFunctionPathSegment segment,
+            CancellationToken cancellationToken, IQueryable bindingParameterValue)
+        {
+            // model build does not support operation with same name
+            // So method with same name but different signature is not considered.
+            MethodInfo method = Api.GetType().GetMethod(segment.Function.Name, BindingFlags.Public | BindingFlags.Instance | BindingFlags.FlattenHierarchy);
+            var parameterArray = method.GetParameters();
+
+            var model = Api.GetModelAsync(cancellationToken).Result;
```
rayao: await

src/Microsoft.Restier.Publishers.OData/Query/RestierQueryBuilder.cs
```diff
@@ -42,6 +42,7 @@ public RestierQueryBuilder(ApiBase api, ODataPath path)
             this.handlers[ODataSegmentKinds.EntitySet] = this.HandleEntitySetPathSegment;
             this.handlers[ODataSegmentKinds.Singleton] = this.HandleSingletonPathSegment;
             this.handlers[ODataSegmentKinds.UnboundFunction] = this.HandleUnboundFunctionPathSegment;
+            this.handlers[ODataSegmentKinds.Function] = this.HandleBoundFunctionPathSegment;
```
lewischeng-ms: Better add a static readonly action in this class probably called `NoOpHandler` whose body contains nothing and set it to both bound and unbound function handlers to make the code less and clear.

src/Microsoft.Restier.Publishers.OData/RestierController.cs
```diff
@@ -76,16 +80,59 @@ private ApiBase Api
                 throw new InvalidOperationException(Resources.ControllerRequiresPath);
             }
 
+            var result = default(IQueryable);
```
lewischeng-ms: Just `IQueryable result;` here.

src/Microsoft.Restier.Publishers.OData/RestierController.cs
```diff
@@ -76,16 +80,59 @@ private ApiBase Api
...
+
+            // Get queryable path builder to builder
             IQueryable queryable = this.GetQuery();
-            QueryRequest queryRequest = new QueryRequest(queryable)
+
+            // If it is a unbound function, call Api function directly
+            ODataPathSegment lastSegment = path.Segments.LastOrDefault();
+            if (lastSegment != null && lastSegment.SegmentKind != ODataSegmentKinds.UnboundFunction)
```
lewischeng-ms: Refactor the code in this method (from line 90 to line 133):
 - Too many condition `lastSegment != null`. Actually if it is null, no result will be created (though I even doubt if this case would happen). Thus either throw exception (if necessary) or just remove this condition.
 - Three if(s) (line 90, line 118 and line 128) can be merged to make the code much clearer. The main logic here has three parts: unbound, bound and the others. Suggest we write a switch statement (on `lastSegment.SegmentKind`) with three cases; extract line 92 to line 114 to a private function. Each case just needs to call two to three methods.

src/Microsoft.Restier.Publishers.OData/RestierController.cs
```diff
@@ -443,20 +490,19 @@ private IQueryable GetQuery()
...
+
+        private IQueryable ApplyQueryOptions(IQueryable queryable, bool applyCount)
+        {
             if (this.shouldWriteRawValue)
             {
                 // Query options don't apply to $value.
                 return queryable;
             }
 
+            ODataPath path = this.GetPath();
```
lewischeng-ms: `this.GetPath` is expensive. Consider pass it through parameter.

src/Microsoft.Restier.Publishers.OData/RestierController.cs
```diff
@@ -525,6 +578,127 @@ private ODataPath GetPath()
             return originalValues;
         }
 
+        private async Task<IQueryable> ExecuteUnboundFunction(UnboundFunctionPathSegment segment,
```
lewischeng-ms: The major part of `ExecuteUnboundFunction` and `ExecuteBoundFunction` is the same. The difference is how we handle the first parameter (the binding parameter). Consider extract the common logic and make it clear, e.g., `GetMethod`, `ReadBindingParameter` (optional for unbound), `ReadParameters`, `ExecuteFunction` (suggest renaming it to more precise `InvokeMethod`).

src/Microsoft.Restier.Publishers.OData/RestierController.cs
```diff
@@ -525,6 +578,127 @@ private ODataPath GetPath()
...
+                {
+                    parameters.Add(bindingParameterValue);
+                }
+            }
+            else
+            {
+                parameters.Add(bindingParameterValue.SingleOrDefault());
+            }
+
+            for (int i=1;i< parameterArray.Length;i++)
```
lewischeng-ms: If we still keep this code, please fix the style.

src/Microsoft.Restier.Publishers.OData/RestierController.cs
```diff
@@ -525,6 +578,127 @@ private ODataPath GetPath()
...
+        {
+            // model build does not support operation with same name
+            // So method with same name but different signature is not considered.
+            MethodInfo method = Api.GetType().GetMethod(segment.Function.Name, BindingFlags.Public | BindingFlags.Instance | BindingFlags.FlattenHierarchy);
+            var parameterArray = method.GetParameters();
+
+            var model = await Api.GetModelAsync(cancellationToken);
+
+            // Parameters of method and model is exactly mapped or there is parsing error
+            var parameters = new Collection<object>();
```
lewischeng-ms: Consider using array directly to avoid future `ToArray()`.

src/Microsoft.Restier.Publishers.OData/RestierController.cs
```diff
@@ -525,6 +578,127 @@ private ODataPath GetPath()
...
+                parameters.Add(convertedValue);
+            }
+
+            return await ExecuteFunction(method, parameters, cancellationToken);
+        }
+
+        private async Task<IQueryable> ExecuteFunction(MethodInfo method, Collection<object> parameters, CancellationToken cancellationToken)
+        {
+            object result = method.Invoke(Api, parameters.ToArray());
+            var model = await Api.GetModelAsync(cancellationToken);
```
lewischeng-ms: `Api.GetModelAsync` is called multiple times during executing a function. Consider cache it to private member or pass through parameter.

Also for the `returnType`. Resolve it from the model takes time. Consider cache it as well.

src/Microsoft.Restier.Publishers.OData/RestierController.cs
```diff
@@ -525,6 +578,127 @@ private ODataPath GetPath()
...
+            IEdmTypeReference returnType = method.ReturnType.GetReturnTypeReference(model);
+            var asQueryableMethod = typeof(Queryable).GetMethod(
```
lewischeng-ms: For these "constant" instances, we should only compute them once and put them into a static class just like `Where`, `Select`, etc.

src/Microsoft.Restier.Publishers.OData/RestierController.cs
```diff
@@ -525,6 +578,127 @@ private ODataPath GetPath()
...
+            if (enumerableType != null)
+            {
+                // This means function is bound to an entity set.
+                // IQueryable should always have generic type argument
+                var elementClrType = enumerableType.GenericTypeArguments[0];
+
+                // For entity set, user can write as ICollection<> or IEnumerable<> or array as method parameters
+                if (parameterArray[0].ParameterType.IsArray)
+                {
+                    var toListMethodInfo = typeof(Enumerable).GetMethod("ToArray").MakeGenericMethod(elementClrType);
```
lewischeng-ms: Also for `ToArray`.

src/Microsoft.Restier.Publishers.OData/RestierController.cs
```diff
@@ -525,6 +578,127 @@ private ODataPath GetPath()
...
+                    var listResult1 = toListMethodInfo.Invoke(null, new object[] { bindingParameterValue });
+                    parameters.Add(listResult1);
+                }
+                else if (parameterArray[0].ParameterType.FindGenericType(typeof (ICollection<>)) != null)
+                {
+                    var toListMethodInfo = typeof (Enumerable).GetMethod("ToList").MakeGenericMethod(elementClrType);
```
lewischeng-ms: Also for `ToList`.

src/Microsoft.Restier.Publishers.OData/RestierController.cs
```diff
@@ -525,6 +578,127 @@ private ODataPath GetPath()
...
+                                        "AsQueryable",
+                                        BindingFlags.Static | BindingFlags.Public,
+                                        null,
+                                        new[] { typeof(IEnumerable<>) },
+                                        null);
+
+            if (returnType.IsCollection())
+            {
+                if (result == null)
+                {
```
lewischeng-ms: Consider extracting the body into a static method `CreateEmptyQueryable(elementType)` (please note the **spell**).

src/Microsoft.Restier.Publishers.OData/RestierController.cs
```diff
@@ -525,6 +578,127 @@ private ODataPath GetPath()
...
+                    var elementClrType = method.ReturnType.GetElementType() ??
+                                         method.ReturnType.GenericTypeArguments[0];
+                    var constructor = typeof(List<>).MakeGenericType(elementClrType).GetConstructor(Type.EmptyTypes);
+                    var instance = constructor.Invoke(new object[] {});
+                    var emptyQuerable = asQueryableMethod.Invoke(null, new object[] { instance }) as IQueryable;
+                    return emptyQuerable;
+                }
+
+                var enumerableType = result.GetType().FindGenericType(typeof(IEnumerable<>));
```
lewischeng-ms: If the result is a `IQueryable`, what's the consequence of calling `AsQueryable`.

src/Microsoft.Restier.Publishers.OData/RestierController.cs
```diff
@@ -525,6 +578,127 @@ private ODataPath GetPath()
...
+                if (enumerableType != null)
+                {
+                    return ((IEnumerable)result).AsQueryable();
+                }
+            }
+
+            // cannot return new[] { result }.AsQueryable(); as need to return in its own type but not object type
+            var objectQueryable = new[] { result }.AsQueryable();
+            var castMethodInfo = typeof(Enumerable).GetMethod("Cast").MakeGenericMethod(method.ReturnType);
```
lewischeng-ms: Also for `Cast`.

src/Microsoft.Restier.Publishers.OData/RestierController.cs
```diff
@@ -525,6 +578,127 @@ private ODataPath GetPath()
...
+
+                var enumerableType = result.GetType().FindGenericType(typeof(IEnumerable<>));
+                if (enumerableType != null)
+                {
+                    return ((IEnumerable)result).AsQueryable();
+                }
+            }
+
+            // cannot return new[] { result }.AsQueryable(); as need to return in its own type but not object type
+            var objectQueryable = new[] { result }.AsQueryable();
```
lewischeng-ms: Hasn't quite got the point here...if control goes here, what does `result` look like? These code looks to me equivalent to `new[] {result}.AsQueryable().Cast<object>().AsQueryable()`...

src/Microsoft.Restier.Publishers.OData/Formatter/Deserialization/DeserializationHelpers.cs
```diff
@@ -0,0 +1,90 @@
...
+using Microsoft.Restier.Core;
+
+namespace Microsoft.Restier.Publishers.OData.Formatter.Deserialization
+{
+    /// <summary>
+    /// Get clr type from payload.
+    /// </summary>
+    internal static class DeserializationHelpers
+    {
+        internal static async Task<object> ConvertValueAsync(object oDataValue, Type expectedReturnType, IEdmTypeReference propertyType, ApiBase api)
```
lewischeng-ms: Do we have a public API to use from Web API OData instead of writing it by ourselves?

chinadragon0515: No, I talked with Sam before I started this part.

src/Microsoft.Restier.Publishers.OData/Query/RestierQueryBuilder.cs
chinadragon0515: Updated

src/Microsoft.Restier.Publishers.OData/RestierController.cs
chinadragon0515: Updated

src/Microsoft.Restier.Publishers.OData/RestierController.cs
chinadragon0515: ODataSegmentKinds cannot be used as case condition, and error is "A constant value is expected", I continue use if but extract logic into different method.

src/Microsoft.Restier.Publishers.OData/RestierController.cs
chinadragon0515: Updated.

src/Microsoft.Restier.Publishers.OData/RestierController.cs
chinadragon0515: Updated

src/Microsoft.Restier.Publishers.OData/RestierController.cs
chinadragon0515: It is new[] {result}.AsQueryable().Cast<Type>().AsQueryable(), we need such convert or apply query will can not find element for object from Model.

src/Microsoft.Restier.Publishers.OData/RestierController.cs
chinadragon0515: Fixed, and moved to a separate method.

src/Microsoft.Restier.Publishers.OData/RestierController.cs
chinadragon0515: Updated, also in DeserializerHelper method

src/Microsoft.Restier.Publishers.OData/RestierController.cs
chinadragon0515: Updated, also in DeserializerHelper method

src/Microsoft.Restier.Publishers.OData/RestierController.cs
chinadragon0515: Updated, also in DeserializerHelper method

src/Microsoft.Restier.Publishers.OData/RestierController.cs
chinadragon0515: Added

src/Microsoft.Restier.Publishers.OData/RestierController.cs
chinadragon0515: Updated

src/Microsoft.Restier.Publishers.OData/RestierController.cs
chinadragon0515: Updated.

src/Microsoft.Restier.Publishers.OData/RestierController.cs
chinadragon0515: Per discussion, no issues.

Issue #432, opened 2016-06-09
RESTier does not actionably handle model building exceptions.


advancedrei:
If you have an entity in your model that leverages a View, OR you simply forget to add a key onto a Code First model, RESTier throws an exception in such a way that the debugger doesn't pause on the offending code. 

### Assemblies Affected
RESTier 0.5.0-beta

### Steps to Reproduce
- Create a simple code-first Data Model. 
- Add an Entity without a key defined.
- Bootstrap your API and try to run it with a debugger attached.

### Expected Result
I would expect that I'd get an exception that can be directly handled, or at least displayed to the end user in a way that gets them the answer quickly. (I believe Issue #436 is an example of this in action).

### Actual Result
A System.AggregateException is being thrown because the GetModelAsync() function is asynchronous, meaning that the real exception is being wrapped. Because the exception is happening really far up the chain, the debugger can't point you to a line of code causing the problem. Because of the examples and documentation, it's not obvious where or how to handle this exception.

chinadragon0515:
Throw the inner exception now, let me know if you have any more suggestions or comments, if not, help to close the issue, thanks.

chinadragon0515:
Close the issue as no response for about 2 weeks. If you have any more suggestions or comments, open new issues. thanks

PR #452, opened 2016-06-23
Improve model built exception been thrown

chinadragon0515:
### Issues
*This pull request fixes issue #432 .*  

### Description
If we use call api.GetModelAsync().Result, then the exception been thrown is wrapped and only inner exception has meaningful message.

If we change to await api.GetModelAsync(), then the exception been thrown will not be wrapped and the exception is meaningful.

### DIFF
--- a/src/Microsoft.Restier.Publishers.OData/Routing/HttpConfigurationExtensions.cs
+++ b/src/Microsoft.Restier.Publishers.OData/Routing/HttpConfigurationExtensions.cs
@@ -51,7 +51,7 @@ namespace Microsoft.Restier.Publishers.OData.Routing
             });
             using (var api = apiFactory())
             {
-                var model = api.GetModelAsync().Result;
+                var model = GetModel(api);
 
                 var conventions = CreateRestierRoutingConventions(config, model, apiFactory);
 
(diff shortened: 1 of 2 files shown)


src/Microsoft.Restier.Publishers.OData/Routing/HttpConfigurationExtensions.cs
```diff
@@ -129,5 +128,29 @@ public static class HttpConfigurationExtensions
...
+        {
+            // Here await is not used because if method MapRestierRoute is mapped async,
+            // Then during application starts, the http service initialization may complete first before this method call is complete.
+            // Then all request will fail, and this happen for some test cases before when get model takes long time.
+            IEdmModel model;
+            try
+            {
+                model = api.GetModelAsync().Result;
+            }
+            catch (Exception e)
```
lewischeng-ms: better use AggreateException and no need to check if it contains an InnerException (because it always has)

lewischeng-ms: http://referencesource.microsoft.com/#mscorlib/system/AggregateException.cs,72

lewischeng-ms:
Looks good to me.

Issue #413, opened 2016-05-27
Exception "EntityType has no key defined" when upgrading to 0.5


danroot:
I have an application using Restier 0.4-rc2 with an EF 6.1.3 dbcontext.  The dbcontext uses external configuration classes to define schema, like this:

     public class AppDbContext{
      ...
       protected override void OnModelCreating(System.Data.Entity.DbModelBuilder modelBuilder)
        {
            base.OnModelCreating(modelBuilder);

            modelBuilder.Configurations.Add(new AttorneyConfiguration());
             ...
       }
     }

    public partial class AttorneyConfiguration : System.Data.Entity.ModelConfiguration.EntityTypeConfiguration<Attorney>
       {
           public AttorneyConfiguration()
               : this("dbo")
           {
           }
 
        public AttorneyConfiguration(string schema)
        {
            ToTable(schema + ".Attorney");
            HasKey(x => x.AttorneyName);
            ....
      }
     }

These classes are generated using the EF Reverse POCO generator.  https://visualstudiogallery.msdn.microsoft.com/ee4fcff9-0c4c-4179-afd9-7a2fb90f5838 

This all works as expected in 0.4 and a working OData endpoint is generated on top of this EF context with routes for all DbSets including Attorney.

Upgrading to 0.5 and switching from DbApi to EntityFrameworkApi, I get the error "EntityType Attorney has no key defined" when calling MapRestierRoute, even though it does in fact have a key defined both in the DB and in the configuration class.   I suspect the Edm generation process is ignoring the EF configuration classes in .5 even though it did not ignore them in .4.  I suspect "Attorney" is mentioned here just because it is the first one MapRestierRoute hits.

Any workaround for this would be appreciated- I'd really like to test some of the other .5 features but this is blocking me.  Is there a way to ensure the Edm model is built using the full DbModel generated by EF?

chinadragon0515:
In 0.5, we change the way to build model, the model is built with Web Api OData conversion model build, it reads all information from CLR class, the way it find the keys by these conversions,
1. Any property defined with Key attribute in CLR class.
2. If no property defined with key attribute, then it will find key by conversion, it will be property with name ID or EntityTypeID (like AttorneyID), case is ignore when finding the attribute.

So for your case, there are two ways to fix the issue,
1. Add key attribute to the AttorneyName property of your class.
2. Or change the name of attorneyName to ID or attorneyID.

Either way will make model build work, let me know if this does not work for you.

[...]

PR #454, opened 2016-06-27
Make model auto pop keys from Entity framework.

chinadragon0515:
### Issues
*This pull request fixes issue #413 .*  

### Description
In Restier 0.5, the model building uses Web APi OData conversion model builder which finds keys of entity type based on ,
a. Whether there is any proeprty marked with key attribute.
b. Whether there is property with name ID or entitytypeNameID
But if the entity type does not define key in this way, then model build will fail

This issue retrieves keys from entity framework and set to model builder.

- [ x ] Test cases added


test/ODataEndToEnd/Microsoft.OData.Service.Sample.Northwind.Tests/Baselines/TestGetNorthwindMetadata.txt
```diff
@@ -12,34 +12,71 @@
         <Property Name="Picture" Type="Edm.Binary" />
         <NavigationProperty Name="Products" Type="Collection(Microsoft.OData.Service.Sample.Northwind.Models.Product)" />
       </EntityType>
-      <EntityType Name="Contact">
+      <EntityType Name="Product">
```
chinadragon0515: Already have detail compare, the changes in this file is only sequence change, no single line content change.

Issue #429, opened 2016-06-08
Make Query related interface method async


chinadragon0515:
Now we have interfaces like IQueryExpressionAuthorizer, IQueryExpressionExpander, IQueryExpressionProcessor, IQueryExpressionSourcer, there are defined synchronized interfaces, which are inconsistent with submit related interfaces.

Change query related interfaces to make them asynchronized.

rayao:
These interfaces are mainly for expression manipulation, maybe async is rarely needed.
But maybe some internal team would welcome it for external data scenarios.

chinadragon0515:
It does not take real effect to make interface async now, close the issue.

PR #463, opened 2016-07-12
Make query related interface async

chinadragon0515:
### Issues
*This pull request fixes issue #429 *  

### Description
Make Query related interface method async to be consistent cross project

- [ x ] Test cases added

### DIFF
--- a/src/Microsoft.Restier.Core/Query/DefaultQueryHandler.cs
+++ b/src/Microsoft.Restier.Core/Query/DefaultQueryHandler.cs
@@ -117,6 +117,11 @@ namespace Microsoft.Restier.Core.Query
 
             public IEdmEntitySet EntitySet { get; private set; }
 
+            /// <summary>
+            /// The method can not be marked as async and return Task as it overwrites super class methods.
+            /// </summary>
+            /// <param name="node">The node to be visited</param>
+            /// <returns>The expression after visitor</returns>
             public override Expression Visit(Expression node)
             {
                 if (node == null)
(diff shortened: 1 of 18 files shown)


src/Microsoft.Restier.Core/Query/DefaultQueryHandler.cs
```diff
@@ -146,17 +151,19 @@ public override Expression Visit(Expression node)
                     }
 
                     // Inspect the visited node
-                    this.Inspect();
+                    this.InspectAsync();
```
rayao: This can return before inspecting actually completes, and so you ignore inspection result.

src/Microsoft.Restier.Core/Query/DefaultQueryHandler.cs
```diff
@@ -146,17 +151,19 @@ public override Expression Visit(Expression node)
...
 
                     // Try to expand the visited node
                     // if it represents API data
                     if (this.context.ModelReference is DataSourceStubModelReference)
                     {
-                        node = this.Expand(visited);
+                        // Can not use await here as the method cannot be async
+                        node = this.ExpandAsync(visited).Result;
```
rayao: These .Result negate the benefits you gain from making those calls async. You need to make the entire call chain async all the way up, or maybe just leave everything as is.

chinadragon0515: I agree with you, I may discard this PR, will discuss with Lewis.

src/Microsoft.Restier.Core/Query/DefaultQueryHandler.cs
chinadragon0515: Yeah, we will need to add .result here, but this makes no sense to make method aync.

chinadragon0515:
After discussion, discard this PR, will revisit whole project to see what we want async and what we want sync.

Issue #469, opened 2016-07-22
Ability to call convention methods from C#


robertmclaws:
We have a complex object model derived from a RESTier service that we are manipulating in a browser. Since Breeze and many other JS frameworks don't properly support OData V4 and/or batching, we want to use an Action to post that complex object to the service. Inside that action, we want to break the object apart and submit it to the different RESTier convention methods in C#, so that each object being inserted can hit the validation logic in OnXXXXInserting, etc. It would basically be a "poor man's batch".

I'm imagining code that looks like this:

``` csharp
    public partial class TestApi : EntityFrameworkApi<CorporationsDbContext>
    {
        public bool SubmitCorporation(Corporation corporation)
        {
            using (var transaction = new TransactionScope())
            {
                'Here is one option for what the function would look like
                this.InsertObject(corporation);
                foreach (var principal in corporation.Principals)
                {
                    'Here is another option
                    this.Execute(EntityFrameworkApiMethods.Insert, principal);
                }
                foreach (var image in corporation.Images)
                {
                    'Here is a third possible option.
                    this.PostObject("Images", image);
                }
                transaction.Complete();
            }
        }
    }
```

Is that something that is possible today? If not, is it something that can get in there? Thanks!

rayao:
OData batch?

chinadragon0515:
If batch does not work for you and you only want to call conversion based method, then based the conversion you want to call like OnInserting and OnInserted, you can do something like this (option 2 in your case)

```
                    var processor = context.GetApiService<IChangeSetItemProcessor>();
                    if (processor != null)
                    {
                        await processor.OnProcessingChangeSetItemAsync(context, item, cancellationToken);
                    }
```

But here you need to construct submit context and ChangeSetItem by your self.
Will this work for you?

advancedrei:
@rayao, very few javascript frameworks properly support OData Batch, so batch won't work for us.

@chinadragon0515, this looks like it could work... How would you construct the submit context? Does the code in your example do that?

Thanks guys!

chinadragon0515:
For SubmitContext, you can refer to class ApiContextExtensions, method SubmitAsync, that's place restier code create SubmitContext,  this the link of code https://github.com/OData/RESTier/blob/master/src/Microsoft.Restier.Core/ApiContextExtensions.cs#L464

[...]

PR #500, opened 2016-08-24
Make the OData Extensions public

robertmclaws:
### Issues
There is no issue open for this request,

### Description
There are situations where this code is useful, for example when having to override some of the built-in behavior to do custom batching. Being able to access extensions like `GetClrType()` is incredibly useful, and making it public would minimize the possibility of duplicate code across codebases.

Also, the team should consider refactoring some of these helpers to live back in the foundational OData classes.

- [   ] Test cases added

### Additional work necessary
- None AFAIK

### DIFF
--- a/src/Microsoft.Restier.Publishers.OData/Extensions.cs
+++ b/src/Microsoft.Restier.Publishers.OData/Extensions.cs
@@ -18,7 +18,7 @@ using Microsoft.Restier.Publishers.OData.Model;
 
 namespace Microsoft.Restier.Publishers.OData
 {
-    internal static class Extensions
+    public static class Extensions
     {
         private const string PropertyNameOfConcurrencyProperties = "ConcurrencyProperties";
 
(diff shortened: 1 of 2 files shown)


chinadragon0515:
Do you need all method in these two class public? If we make the class public, I will need to revisit all the method name and so on. And then have internal review. This will take some time.

robertmclaws:
I mean, you don't HAVE to, if you create a way that lets you submit entities through the Restier context on the server side, like I mention in #469. Right now it takes > 200 lines of code and reflection on private methods to make it happen. Should be vastly simpler.

chinadragon0515:
@robertmclaws  now we focus on the protocol support with limit resource, for the items which are not defined by protocol, we do not have a plan to support, and we accept contributions which could become protocol. 

Can I know which method you need now to be public so I can only public that methods first.

chinadragon0515:
WIll make method GetTypeReference and GetClrType enough for you? For others, should be internal logic, plan to keep as internal now.
If yes, a new class will be created and have the public method.

robertmclaws:
I think that would be fine, yes.

chinadragon0515:
OK, I will close this PR, and will create a new PR to cover this, thanks.

Issue #505, opened 2016-09-01
PATCH semantics against complex types is incorrect


mikepizzo:
PATCH against a complex typed property is currently being handled with PUT semantics -- properties not specified in the payload are being set to null, where they should be left untouched.

### Reproduce steps
PATCH http://services.odata.org/TripPinRESTierService/(S(h0jbxldzeiopthtrxz5zzme5))/Airports('KLAX')
{
    "Location":{
    "Address":"1 World Way, Los Angeles, CA, 90045","City":{
      "Region":"CA"
    }
  }
}

### Expected result
Only the Region of the City property should be changed to "CA"; the rest of the payload should remain unchanged.
 
### Actual result
The other properties of Location are set to null:
{
"@odata.context":"http://services.odata.org/TripPinRESTierService/$metadata#Airports/$entity",
"Name":"Los Angeles International Airport",
 "IcaoCode":"KLAX",
 "IataCode":"LAX",
  "Location":{
    "Address":"1 World Way, Los Angeles, CA, 90045",
     "City":{
        "Name":null,
        "CountryRegion":null,
        "Region":"CA"
    },
    "Loc":null
  }
}
 

### Additional details
PATCH http://services.odata.org/TripPinRESTierService/(S(h0jbxldzeiopthtrxz5zzme5))/Airports('KLAX')
{
   "Name":"Los Angeles International Airport"
}
Correctly only updates the Name property; the problem only appears if the payload contains a subset of the properties of a complex type.

chinadragon0515:
@mirsking I merge the PR to build new release. Help to update the hosted sample services on odata.org and sample services in odatasamples git repository.

chinadragon0515:
Sample is updated, close the issue.

PR #507, opened 2016-09-01
fix issue 505 and add patch test cases, also support key as segment and add CORS header

mirsking:
### Issues
*This pull request fixes issue #505, also support key as segment and add CORS header.*  

### Description
Complex type property now can be patched rightly.

- [ x ] Test cases added

### DIFF
--- a/src/Microsoft.Restier.Providers.EntityFramework/Submit/ChangeSetInitializer.cs
+++ b/src/Microsoft.Restier.Providers.EntityFramework/Submit/ChangeSetInitializer.cs
@@ -220,7 +220,7 @@ namespace Microsoft.Restier.Providers.EntityFramework
                                 propertyPair.Key));
                         }
 
-                        value = Activator.CreateInstance(type);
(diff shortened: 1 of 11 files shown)


src/Microsoft.Restier.Providers.EntityFramework/Submit/ChangeSetInitializer.cs
```diff
@@ -220,7 +220,7 @@ private void SetValues(DbEntityEntry dbEntry, DataModificationItem item, Type re
                                 propertyPair.Key));
                         }
 
-                        value = Activator.CreateInstance(type);
+                        value = propertyEntry.CurrentValue;
                         SetValues(value, type, dic);
```
chinadragon0515: I am OK for this changes.
But add a TODO to line 257, the SetValues method is called by both patch and patch, we need to consider complex property has other complex type property cases.

Open an issue to track this and add to that TODO.

Issue #527, opened 2016-09-27
Restier cannot handle more than one API registered per app.


robertmclaws:
It appears that the design for Restier did not take into account the possibility that more than one API could be registered with inside a single project. 
### Assemblies affected

RESTier 1.0.0-beta  
### Reproduce steps

Register 2 APIs in a single project using MapRestierRoute.
### Expected result

I would expect a request to the service to return normally.
### Actual result

The Dependency Injection system is throwing an error because there is more than one model registered, but the call is apparently calling the LINQ `.Single()` function, which only expects one element. It should likely be calling `.FirstOrDefault()` instead and checking for a null result.

``` js
{
  "Message": "An error has occurred.",
  "ExceptionMessage": "One or more errors occurred.",
  "ExceptionType": "System.AggregateException",
[...]

PR #529, opened 2016-09-28
Fix ModelProducer to handle multiple APIs with different DbContexts

robertmclaws:
### Issues

This pull request fixes issue #527.  
### Description

New code checks the collection of objects returned by `efModel.GetItems<EntityContainer>(DataSpace.CSpace)` for the `EntityContainer` that matches the expected `DbContext`. If none is found, throws an informative Exception that helps the user identify the problem.
- [ ] Test cases added
- [ ] Build and test with one-click build and test script passed
### Additional work necessary

Proper tests need to be added to make sure that GetModelAsync() works in various situations. There is not enough direct code coverage for that method.


src/Microsoft.Restier.Providers.EntityFramework/Model/ModelProducer.cs
```diff
@@ -62,7 +62,19 @@ public Task<IEdmModel> GetModelAsync(ModelContext context, CancellationToken can
...
             var efModel = (dbContext as IObjectContextAdapter).ObjectContext.MetadataWorkspace;
-            var efEntityContainer = efModel.GetItems<EntityContainer>(DataSpace.CSpace).Single();
+            var efEntityContainers = efModel.GetItems<EntityContainer>(DataSpace.CSpace);
+            var efEntityContainer = efEntityContainers.FirstOrDefault(c => c.Name == dbContext.GetType().Name);
+            if (efEntityContainer == null)
+            {
+                if (efEntityContainers.Count > 1)
+                {
+                    var containerNames = efEntityContainers.Aggregate("", (current, next) => next.Name + ", ");
+                    throw new Exception("This project has multiple EntityFrameworkApis using different DbContexts, and the correct contect could not be loaded. \r\n" +
```
rayao: spell: contect -> context
I'm not sure if C# 6 syntax could pass compile, Vincent?

robertmclaws: OK, I fixed the spelling issue. Thanks for catching that.

I'm not sure how you are compiling, but if you're on VS2015 it should be fine. Can always do a string.Format() if necessary.

chinadragon0515: Our internal signed build server is still VS2013, we can not use c# 6 syntax.

src/Microsoft.Restier.Providers.EntityFramework/Model/ModelProducer.cs
```diff
@@ -62,7 +62,19 @@ public Task<IEdmModel> GetModelAsync(ModelContext context, CancellationToken can
             var dbContext = context.GetApiService<DbContext>();
 
             var efModel = (dbContext as IObjectContextAdapter).ObjectContext.MetadataWorkspace;
-            var efEntityContainer = efModel.GetItems<EntityContainer>(DataSpace.CSpace).Single();
+            var efEntityContainers = efModel.GetItems<EntityContainer>(DataSpace.CSpace);
+            var efEntityContainer = efEntityContainers.FirstOrDefault(c => c.Name == dbContext.GetType().Name);
+            if (efEntityContainer == null)
+            {
+                if (efEntityContainers.Count > 1)
+                {
```
chinadragon0515: Do we know which case we will have two container, what will each container will have? If we fix like this, and EF do have two containers, each container will have some entity set (DbSet), then the logic will be wrong, we will need to iterate the container to get the information.

I am OK to add this as a temp fix but please help to add a TODO comment here for further improvement.

src/Microsoft.Restier.Providers.EntityFramework/Model/ModelProducer.cs
```diff
@@ -62,7 +62,19 @@ public Task<IEdmModel> GetModelAsync(ModelContext context, CancellationToken can
...
+                    var containerNames = efEntityContainers.Aggregate("", (current, next) => next.Name + ", ");
+                    throw new Exception("This project has multiple EntityFrameworkApis using different DbContexts, and the correct context could not be loaded. \r\n" +
+                        $"The contexts available are '{containerNames.Substring(0, containerNames.Length - 2)}' but the Container expects '{efEntityContainer.Name}'.");
+                }
```
chinadragon0515: The exception string is not hard coded, we need to put into property files, refer to other exception for sample.

src/Microsoft.Restier.Providers.EntityFramework/Model/ModelProducer.cs
robertmclaws: Yes, I know which case there will be more than one container, that's why I had to write this code. If you register 2 APIs with 2 different DB contexts in the same website, then the call to `efModel.GetItems<EntityContainer>(DataSpace.CSpace)` will return more than one container, even though they are in separate EDMX files in separate DbContexts. One would think that wouldn't be the case, so maybe this exposed a flaw somewhere deeper in the OData or EF framework. But that is OK, because the code above these lines *should* always return the right container for the particular context. This particular set of highlighted code is only there in case, for some presently unknown reason, the query to get the matching context fails, It's a defensive measure to help developers understand what is going on, instead of the current code just assuming everything will work, until it doesn't, and not giving you any clue as to why.

src/Microsoft.Restier.Providers.EntityFramework/Model/ModelProducer.cs
robertmclaws: I hard-coded the string for now so that you could see the exception that is thrown, and there can be a possible discussion about the language, if necessary. I also wanted the team to see how to throw an informative exception (most of the exceptions in Restier are not very informative). We can move the strings into the property files after everyone is OIK with it.

src/Microsoft.Restier.Providers.EntityFramework/Model/ModelProducer.cs
robertmclaws: Is it possible to upgrade the build server? It seems like the VS2015 compiler with all of the updates would be able to take advantage of more optimizations / compilation improvements than a 4 year old runtime. If this project is going to be moved to be community-driven, people are not going to want to deal with legacy systems.

chinadragon0515: I do not think it will happen in short time consider the server is maintained and used by much wide d groups. We do already initiate such discussion

robertmclaws: Alright, I'll take it out when we move the text into a Resource.

src/Microsoft.Restier.Providers.EntityFramework/Model/ModelProducer.cs
robertmclaws: I believe that this code would work even if an individual EDMX file ever had multiple containers (though I don't know if that is actually even possible). Given the fact that right now, the code is pulling containers from totally separate `DbContext` instances and putting them in the same collection, one would think that multiple Containers in the same Context would also apply. However, if at some point in the future it didn't, the logic for `efEntityContainers` could be adjusted easily; and that's why I broke that call out from the `.FirstOrDefault()` call.

chinadragon0515: OK, thanks. Can you help to add some comments here which will help people who maintain these codes.

src/Microsoft.Restier.Providers.EntityFramework/Model/ModelProducer.cs
chinadragon0515: For the message part, you are native speaker, you are much better than me on this :)
BTW, as I will start vacation tomorrow, I will response to this PR after my vacation.

src/Microsoft.Restier.Providers.EntityFramework/Model/ModelProducer.cs
robertmclaws: Done. Hopefully those help. I will see about moving the exception messages to the resources over the weekend.

robertmclaws:
Hey everyone,

Just wanted to see where we're at on this. Thanks!

chinadragon0515:
I am not sure whether you have a chance to move message to resource file, after you have done this, I will merge the changes, but we do not have a solid plan for next release now.

robertmclaws:
Sorry for the delay in this, @chinadragon0515. I've made the changes so that now the Exception uses the Resources instead of a hard-coded string. Ideally, this situation still needs a test case, and at some point I will endeavor to write one. But for now, we should be able to merge this change,

chinadragon0515:
Merge the PR

PR #560, opened 2017-04-22
Dependency update

robertmclaws:
### Description
- This check-in updates dependencies to their very latest versions, and fixes broken unit tests.
- Also updates the NuGet packaging to generate debug symbol packages.
- Fixed a deadlock in RestierController.cs
- Throws a more informative exception when you hit #564.
- Allows Interceptors to be specified `protected internal` so that they can be called directly from unit tests. 

- [x] Test cases added
- [x] Build and test with one-click build and test script passed

### Additional work necessary
None. Code is ready for integration and to be shipped.


fileman:
Hi, I've forked master branch and updated nuget ref and also changed the target framework to net462, why have you leave it at net45, problem whit net462 or simply you haven't changed?

robertmclaws:
I didn't change it, it's always been at net45. There is nothing wrong with keeping it at net45 either, AFAIK.

adalbus:
I would like to use Restier in our project, but it has a hard dependency to an older version of the DependencyInjection library.

Is there any way to help to expedite this change to solve this issue? Perhaps with a code review?

Disclaimer: I work for my Microsoft.

src/Microsoft.Restier.Publishers.OData/RestierController.cs
```diff
@@ -525,19 +527,25 @@ private IQueryable GetQuery(ODataPath path)
...
 
             if (this.shouldWriteRawValue)
             {
                 // Query options don't apply to $value.
-                return queryable;
+                return (queryable, null);
             }
 
             HttpRequestMessageProperties properties = this.Request.ODataProperties();
-            var model = Api.GetModelAsync().Result;
```
rayao: Actually this won't deadlock since at this time EdmModel must have been ready. It's been ready in WebApiConfig. But of course I welcome this change.

robertmclaws:
@adalbus, I have a fork of this project on NuGet, which you can install by calling `install-package AdvancedREI.Restier -pre`. The latest version has a more flexible dependency on the DependencyInjection library.

HTH!

xuzhg:
@robertmclaws Can you make the codes passed in vs2015?

robertmclaws:
@xuzhg I didn't change any project versions or anything, this should still compile in VS2015 as well.

xuzhg:
@robertmclaws 

Can you help me understand codes like  `Task<(IQueryable Queryable, ETag Etag)> ` can be compiled at VS2015?

robertmclaws:
I thought ValueTuples compiled on VS2015, but I guess they don't that way. I may have to change the syntax.

Have you guys not moved to VS2017 yet?

xuzhg:
@robertmclaws 

We'd love to. But not yet. 
Please let me know if you are ready.

robertmclaws:
@xuzhg, I just pushed an update to this PR that adds the latest Roslyn compilers, and have verified that the code will compile in VS2015. The editor will complain if you open RestierController.cs in the IDE, but you can ignore those errors.

For everyone's understanding, this syntax `Task<(IQueryable Queryable, ETag Etag)>` creates a `ValueTuple`, which is a `Tuple` with names properties, instead of `Item1`, `Item2`, etc. Makes it a little easier to pass around values when there is no need to create a separate type just for marshalling data around.

At any rate, the latest change will let VS2015 and VS2017 users contribute to the project without issue.

bastienlemaitre:
Any news ?

robertmclaws:
I'm closing this pull request because we handled these fixes in https://github.com/OData/RESTier/commit/aabea3c5036f43a8df929154723722dbb873297f

Issue #578, opened 2017-06-27
System.InvalidCastException when using a composite key with a DateTime


0xced:
Entities having a composite key with a DateTime can't be accessed by id.

### Assemblies affected
RESTier 1.0.0-beta

### Reproduce steps
Use RESTier with `EntityFrameworkApi<T>` where the DbContext contains an entity having a composite key with int + DateTime.
Access the entity with `http://localhost/MyService/MyRoute/MyEntities(SomeId=1,SomeDate=2017-06-27)`

### Expected result
Accessing an entity by its key should return the entity.

### Actual result
An internal server error (500) is returned because a `System.InvalidCastException` is thrown.

### Additional details

This error occurs because `SomeDate` is decoded as `DateTimeOffset` and it can't be converted to a `DateTime`.

I should probably make RESTier interpret `SomeDate` as `DateTime` but I have no idea how to achieve this.

Here is the full error which occurs at [CreateEqualsExpression](https://github.com/OData/RESTier/blob/1.0.0-beta/src/Microsoft.Restier.Publishers.OData/Query/RestierQueryBuilder.cs#L123):

```json
{
  "error": {
    "code": "",
    "message": "An error has occurred.",
    "innererror": {
      "message": "Object must implement IConvertible.",
      "type": "System.InvalidCastException",
[...]

PR #584, opened 2017-07-11
Convert DateTimeOffset to DateTime when necessary

0xced:
### Issues
This pull request fixes issue #578.

### Description
Entities with a `DateTime` property in their key can't be created/read/updated/deleted because OData Web API does not support `DateTime` in the keys. This pull request addresses this issue by converting `DateTimeOffset` to `DateTime` when necessary.

### Checklist
- [ ] Test cases added
- [ ] Build and test with one-click build and test script passed

### Additional work necessary
No additional work necessary.

### DIFF
--- a/src/Microsoft.Restier.AspNet/Query/RestierQueryBuilder.cs
+++ b/src/Microsoft.Restier.AspNet/Query/RestierQueryBuilder.cs
@@ -118,7 +118,7 @@ namespace Microsoft.Restier.AspNet.Query
         {
             var property = Expression.Property(parameterExpression, propertyName);
             var constant = Expression.Constant(
-                Convert.ChangeType(propertyValue, property.Type, CultureInfo.InvariantCulture));
+                TypeConverter.ChangeType(propertyValue, property.Type, CultureInfo.InvariantCulture));
             return Expression.Equal(property, constant);
         }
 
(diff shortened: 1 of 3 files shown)


xuzhg:
@0xced Would you please add several tests to cover your change-sets?

rayao:
ChangeSetItem and RestierQueryBuilder are at 2 different level, one is in Core and the other publisher, we'd better make the change more consistent.

0xced:
> Would you please add several tests to cover your change-sets?

@xuzhg As I explained, this problem only exists when a `DateTime` is part of the primary key and I could not find a part of the sample databases used in the tests with a `DateTime` as part of the primary key. If there’s one I could probably write a test but I don't feel like creating a whole new database just for this. :-/

> we'd better make the change more consistent

@rayao What would you suggest to make the change more consistent? I thought that extracting the logic in a helper class was precisely making it consistent.

0xced:
Ping.

rayao:
I'm ok with this change.

robertmclaws:
@0xced I'd like to be able to accept this PR, but you're going to need to catch up on your branch before I can pull it in. Thanks!

0xced:
I just rebased on master. It seems the continuous integration has failed but unfortunately, I'm not authorized to see what has failed.

robertmclaws:
Don't worry about the CI, I broke the project by moving it to an entirely new structure, so it's going to keep failing until Microsoft gives me access to fix things. I just wanted to make sure the code was current and still necessary so I could pull it in. Approving now, and I hope to have a CI build you can you before the end of the weekend. Thanks for your contribution!

PR #586, opened 2017-07-12
Enabling code analysis for RESTier

AlanWong-MS:
- Enabling code analysis during builds by default
- Fixing error regarding unused parameter by removing unused parameter
- Suppressing errors regarding un-instantiated classes because they are actually being referenced and will be utilized during runtime per service call
- Suppressing error regarding marking function as static as the function neither resides in a utility class, nor is appropriate to be static

### DIFF
--- a/src/Microsoft.Restier.Core/Microsoft.Restier.Core.csproj
+++ b/src/Microsoft.Restier.Core/Microsoft.Restier.Core.csproj
@@ -8,7 +8,7 @@
     <RootNamespace>Microsoft.Restier.Core</RootNamespace>
     <AssemblyName>Microsoft.Restier.Core</AssemblyName>
     <DocumentationFile>$(OutputPath)$(AssemblyName).xml</DocumentationFile>
-    <RunCodeAnalysis>$(CodeAnalysis)</RunCodeAnalysis>
+    <RunCodeAnalysis>true</RunCodeAnalysis>
     <CodeAnalysisRuleSet>..\Strict.ruleset</CodeAnalysisRuleSet>
     <!-- TODO: compile as a portable assembly -->
   </PropertyGroup>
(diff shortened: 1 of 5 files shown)


src/Microsoft.Restier.Core/Microsoft.Restier.Core.csproj
```diff
@@ -8,7 +8,7 @@
     <RootNamespace>Microsoft.Restier.Core</RootNamespace>
     <AssemblyName>Microsoft.Restier.Core</AssemblyName>
     <DocumentationFile>$(OutputPath)$(AssemblyName).xml</DocumentationFile>
-    <RunCodeAnalysis>$(CodeAnalysis)</RunCodeAnalysis>
+    <RunCodeAnalysis>true</RunCodeAnalysis>
```
AlanWong-MS: Per discussion with Biao, I am considering changing this line to

<RunCodeAnalysis Condition="'$(RunCodeAnalysis)' == ''">true</RunCodeAnalysis>

xuzhg: +1

robertmclaws:
You guys should really check out my pull requests before you get too deep into Restier changes... I made quite a few fixes to unit testing and code analysis.

robward-ms:
@robertmclaws - Yep, thanks. We chatted about that this morning and will plan to look at all RESTier PRs together, including all those you have contributed.

robertmclaws:
Hey everyone! I'm going to go ahead and close this down... we're handling Code Analysis through the Microsoft.CodeQuality.Analyzers package now. I'll keep an eye on the suppressions that were made and make sure they get ported over to the new codebase. Thanks for the contribution!

PR #587, opened 2017-07-13
Clean the Microsoft.Reliability

xuzhg:
### Issues

Cleanup the Microsoft.Reliability

### Description
Cleanup the Microsoft.Reliability

- [  ] Test cases added


robertmclaws:
Thank you for this PR... I think we're going to look at  re-evaluating ALL of the exclusions before RTM, so I'm going to go ahead and close this for now. But thanks for all your hard work! :)

Issue #629, opened 2019-01-03
Properly implement Constructor Injection for the DI pattern


robertmclaws:
Currently, the product passes around IServiceContainers between classes and grabs whatever it needs whenever it needs it.

However, in order for the Restier architecture to be clean and thoroughly testable, it needs to properly implement Constructor-based Dependency Injection across the entire product surface. This will require a major refactor of the application, which is not likely to happen in the 1.0 timeframe. However, some things can be done in the interim to smooth the transition to that architecture.

jspuij:
Hi @robertmclaws. I've created a pull request for a first series of fixes that addresses this issue. It's PR #645. I've come across a lot of other issues that need a bit more discussion. I did not want to pollute the issues in this repository, so for now these issues are here:

https://github.com/jspuij/restier/issues

If you want me to move those issues over, I will. If you want to discuss them on something like Slack or Teams, i'll wait till that is setup.

The amount of classes that include a reference to Microsoft.Extensions.DependencyInjection has reduced from 37 to 25. All classes that remain are either unit tests, or part of discussion about the design.

robertmclaws:
I think we've gotten as far as we can in this release, so I'm going to go ahead and close this issue. If you think there is anything from your fix branch that *needs* to make it in before RTM, please let me know and we can talk about it.

Thanks!

PR #645, opened 2019-07-19
Part of Constructor Injection for the DI pattern.

jspuij:
### Issues
This pull request fixes part of https://github.com/OData/RESTier/issues/629. 

### Description
These are the quick wins that could be done without major discussion about redesign of certain classes. It also does not change the public API of Restier. All unit tests still pass.
The following sub-issues of https://github.com/OData/RESTier/issues/629 are fixed by this PR:

https://github.com/jspuij/RESTier/issues/1
https://github.com/jspuij/RESTier/issues/5
https://github.com/jspuij/RESTier/issues/6
https://github.com/jspuij/RESTier/issues/9
https://github.com/jspuij/RESTier/issues/10
https://github.com/jspuij/RESTier/issues/16
https://github.com/jspuij/RESTier/issues/17
https://github.com/jspuij/RESTier/issues/21
https://github.com/jspuij/RESTier/issues/22
https://github.com/jspuij/RESTier/issues/23
https://github.com/jspuij/RESTier/issues/25
https://github.com/jspuij/RESTier/issues/26
https://github.com/jspuij/RESTier/issues/27 (Partial, needs a bit of discussion).

- [ x] *Test cases added*
- [ x] *Build and test with one-click build and test script passed*


src/Microsoft.Restier.AspNet/RestierController.cs
```diff
@@ -41,24 +41,91 @@ public class RestierController : ODataController
...
+        /// Resolves the api, query settings, validation settings, operation executor from
+        /// the Request container associated with the HttpRequestMessage.
+        /// </remarks>
+        /// <param name="controllerContext"> 
+        /// The <see cref="HttpControllerContext"/> object that is used for the initialization.
+        /// </param>
+        protected override void Initialize(HttpControllerContext controllerContext)
         {
-            get
+            base.Initialize(controllerContext);
```
jspuij: Move checks and return if all constructor arguments are provided.

src/Microsoft.Restier.EntityFramework/IDbContextProvider.cs
```diff
@@ -0,0 +1,20 @@
...
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Microsoft.Restier.EntityFramework
+{
+    /// <summary>
+    /// Interface to implement by Api classes to get the context.
+    /// </summary>
+    public interface IDbContextProvider
```
robertmclaws: Interfaces are concepts that imply a contract that will be available over an extended period of time. I do not anticipate us supporting this contract in v2, so I don't think we should include it here. If you could please undo the changes in this commit, we can talk later about what this would look like in v2. Thanks!

jspuij: Don't think it's ideal either. However, without an interface, it's impossible to get the DB context as EntityFrameworkApi is generic and you don't know the concrete type. I'm trying to get rid of the "pass the Service provider around and resolve from it everywhere".
There is logic to an interface: If the ApiBase is an EntityFrameworkApi<T> instance, it must be able to provide a DB context. This will be true now, but for every version in the future. We can discuss about whether it should be a method or property,  but you'll need a DB Context in every Filter or Authorizer method you add to your EntityFrameworkApi descendant.

robertmclaws: I had a more long-winded answer for this, but the short version is that DbContext already implements an interface, and I think we can register the context differently to achieve the same goals without introducing an unnecessary interface. We shouldn't be introducing any new contracts at this stage, we're too close to shipping. If you can back this out for now, we can talk about if we're going to fix it before we ship V1.

jspuij: Done. This means that I will have to find a different solution to GetApiService for the DBContext. Or we postpone this until V2.

PR #648, opened 2019-09-24
Reverted to one ServicesCollection again

jspuij:
on the RestierContainerBuilder and added support for pre-odata and post-odata service registration hooks (as constructor arguments).

All unit tests pass.

### DIFF
--- a/src/Microsoft.Restier.Core/RestierContainerBuilder.cs
+++ b/src/Microsoft.Restier.Core/RestierContainerBuilder.cs
@@ -24,7 +24,7 @@ namespace Microsoft.Restier.Core
 
         private readonly Type apiType;
 
-        private readonly Action<IServiceCollection> postOdataConfigureAction;
+        private readonly Action<IServiceCollection> configureAction;
 
         #endregion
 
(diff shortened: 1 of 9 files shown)


src/Microsoft.Restier.Core/RestierContainerBuilder.cs
```diff
@@ -112,11 +115,8 @@ public IContainerBuilder AddService(ODataServiceLifetime lifetime, Type serviceT
         /// <returns>The container built by this builder.</returns>
         public virtual IServiceProvider BuildContainer()
         {
-            foreach (var descriptor in RestierServices)
-            {
-                Services.Add(descriptor);
-            }
-            AddRestierService(!RestierServices.Any());
+            postOdataConfigureAction?.Invoke(Services);
+            AddRestierService(!Services.Any(x => x.ServiceType == typeof(ApiBase)));
```
jspuij: This kind of changes the check, but I think adding the RestierService when no apiBase services have been registered on the services collection before is a sensible one.

robertmclaws:
This feels like the build should have failed, because it didn't update the Sample service to show how interacting with it would work (or at least have a pre- and post- version commented out.

I'm also not sure that we need to expose people to the implementation detail of the order that OData needs to be registered in.

Can you please update the sample to show what it would look like?

jspuij:
Actually it does not fail because I've kept the pre-odata delegate null in UseRestier:

```
var builder = new RestierContainerBuilder(typeof(TApi), null, (services) =>
                {
                    services
                   .AddCoreServices(typeof(TApi))
                   .AddConventionBasedServices(typeof(TApi));

                configureAction(builder.RestierServices);
                    configureAction(services);

                builder.RestierServices.AddRestierServices<TApi>();
                    services.AddRestierServices<TApi>();
                });
```

This creates a new containerbuilder with only a postOdata services registration hook as the third constructor argument. The second one is null. It's not strictly a required delegate. You could just start adding services to the RestierContainerBuilder after it's instantiated, and these services would be added to the RestierContainerBuilder before the Odata services. But I thought RestierContainerBuilder would be more verbose this way.

I'll update the sample to add an UseRestier overload with pre-odata and post-odata registration.

PR #650, opened 2019-09-25
More work on service configuration.

jspuij:
### Issues

Part of https://github.com/OData/RESTier/pull/649

### Description

After a more thorough check of service registration I've got some more changes.

### Additional work necessary

Cloudnimble needs to be updated.


robertmclaws:
Looks good!

PR #682, opened 2021-05-15
Added support for Entity Framework Core.

jspuij:
As discussed as well. Please do not merge yet. This is now based on the core branch, but I'll rebase on master once the core pull request is merged.


jspuij:
OK, So I have rebased it.  It should be OK now. What I have done is moved the shared code (that is mostly equal, just some ifdefs) to a shared project. These shared projects are used by Odata itself as well. The code for EF Core is in a different namespace now, because theoretically, you can use both Entity Framework and EF Core together on .NET core.

PR #703, opened 2022-03-05
Added test to validate batch request/response payloads

mikepizzo:
### Description
*This pull request fixes issue adds a test to validate batch request/response payloads.*  

- [ x ] *Test cases added*
- [ x ] *Build and test with one-click build and test script passed*

### DIFF
--- a/src/Microsoft.Restier.Tests.AspNet/FeatureTests/BatchTests.cs
+++ b/src/Microsoft.Restier.Tests.AspNet/FeatureTests/BatchTests.cs
@@ -10,6 +10,9 @@ using System.Globalization;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Restier.Tests.Shared;
 using System.Threading;
+using System.Net.Mime;
+using System.Net.Http.Headers;
+using System.Text;
 
 #if NETCOREAPP3_1_OR_GREATER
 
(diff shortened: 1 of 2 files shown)


robertmclaws:
`MimePayloadTest` is failing for me on AspNet Classic, and all the batching tests are still failing for me on every AspNetCore permutation. Does something need to be configured differently?

mikepizzo:
I'll take a look. Are they failing if you run individually or just if you run together with the other tests?

robertmclaws:
AspNetCore is failing individually and separately. MimePayloadTest on AspNet Classic passed by itself.

mikepizzo:
Yep; MimePayloadTest is failing when run in after JsonPayloadTest because the data isn't reset between tests so the inserts fail because the records already exist. I think we need to change the test server to reset each time, as I notice a bunch of other tests failing when run in sequence but passing when run individually.

The Aspnetcore batch tests were failing on the initial GET because the baseUri was not set properly for those tests. I've fixed that, but am now getting a bad format error on all of the batch tests (not just the ones I added). I don't think that has anything to do with my changes, but haven't had a chance to track whether that's due to the way the aspnetcore test service is set up or the aspnetcore code itself. Will try and get dig deeper this week.

Issue #709, opened 2022-04-11
Batch requests are still failing on .NET Core 3.1 +


robertmclaws:
The batch unit tests are failing on .NET Core 3.1 and later. Right now, the issue is "unsupported media type" coming back from both XML and JSON posts. If we can get this fixed, we go from 40 failed tests to 28, and I believe we can finally get this thing to 1.0 GA.

robertmclaws:
@caldwell0414 @jspuij any chance you gents could pitch in on figuring this out? The tests are calling the right URLs now, we just need to make sure the processing pipeline is configured properly.

mikepizzo:
First issue is that, for .NET Core, we need to add the ODataBatch middleware. This must be done before calling UseMvc on the api builder.

In RestierBreakdanceTestBase, add builder.UseODataBatching() to line 78, immediately before useMvc:

                builder.UseDeveloperExceptionPage();
                **builder.UseODataBatching();**
                builder.UseMvc(routeBuilder =>
                ...

The second issue, with the JSON tests, is that in .NET Core we return a slightly different content-type within the individual responses (we omit the odata.stream=true for the non-.NET Core). Not sure why this discrepancy, but both are valid so I made allowances in the test.

The final issue is that we appear to have a race condition with the way we process the changeset, such that some entries are not getting properly processed. I'm working on trying to track that one down.

PR #710, opened 2022-04-26
Fix Batch on .NET Core

mikepizzo:
### Issues
*This pull request fixes issue #709.*  

### Description

Issue #1: Batch requests were failing for .NET Core because .NET Core requires the ODataBatch middleware to be registered (must be before UseMvc). It would be nice to do this automatically, but it wasn't clear where we would have access to the apibuilder in order to add the batch middleware. For now, I added the UseODataBatch to RestierBreakdanceTestBase, and developers using RESTier will have to add the same UseODataBatch in order to enable batch support in .NET Core.

Issue #2: In the JSON format, the content-type of batches within a response include odata.streaming=true in .NET Core, but not net472. Not sure why this difference, but both are valid so I just accounted for this in the tests. 

Issue #3: There were some threading issues with how we handled the Changeset entries.  First, entries were kept in a List<T>, which is not threadsafe, although we had multiple threads adding to the list, so I made it a ConcurrentQueue.  Second, and harder to track down, was the fact that we were lazy-instantiating the collection (list or queue) underlying the entries for the changeset.  This was leading to a race condition where one thread would check to see if it existed and, since it didn't, would create it. but before that creation completed, the second thread would see it didn't exist and also create it, but the new one wouldn't include the entry from the first thread. So, I moved creation of the collection (ConcurrentQueue) to the constructors.

- [x] *Test cases added*

### Additional work necessary
none

### DIFF
--- a/src/Microsoft.Restier.AspNet/RestierController.cs
+++ b/src/Microsoft.Restier.AspNet/RestierController.cs
@@ -239,13 +239,13 @@ namespace Microsoft.Restier.AspNet
             if (changeSetProperty is null)
             {
                 var changeSet = new ChangeSet();
-                changeSet.Entries.Add(postItem);
+                changeSet.Entries.Enqueue(postItem);
 
                 var result = await api.SubmitAsync(changeSet, cancellationToken).ConfigureAwait(false);
             }
             else
             {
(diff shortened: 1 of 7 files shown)


robertmclaws (approved):
Looks amazing, great job!

PR #731, opened 2023-10-16
Endpoint routing

jspuij:
### Description
This pull request adds endpoint routing support for Restier with OData 7.4+ on aspnet.core.

- [x] *Test cases added*

### Additional work necessary
I'm not sure about the syntax for AddEndpointRestier instead of AddRestier. This might need a bit of discussion.


robertmclaws:
Hey @jspuij! Just so you are aware, this change required a much more intensive refactor than you might have anticipated. As I started to expand the tests, I encountered a fundamental limitation with how RouteNames are handled in Endpoint Routing, which required us to expand Endpoint Routing support to the full test suite.

That meant I had to make changes to your API design, but I think the end result meshes much more closely with developer expectations. You can follow that work here:

https://github.com/OData/RESTier/pull/738/files

PR #738, opened 2023-11-18
Refactoring the Endpoint Routing API

robertmclaws:
- Rename a bunch of classes (they might break people using reflection but should not be a big deal).
- Refactor `AddRestier()` so that handles both Legacy and Endpoint routing.
- Call `AddControllers()` in both cases because the services involved are minimal and people will expect similar behavior.
- Make the return values consistent so we can reduce complexity on the test classes.
- Still need to refactor the test suite to run the FULL GAMUT of tests on ASP.NET Core for both types of routing. That will be fun to figure out.

- [x] *Test cases added*

### Additional work necessary
*If documentation update is needed, please add "Docs Needed" label to the issue and provide details about the required document change in the issue.*


src/Microsoft.Restier.AspNetCore/Extensions/Restier_IServiceCollectionExtensions.cs
```diff
@@ -115,67 +120,24 @@ public static IMvcBuilder AddRestier(this IServiceCollection services, Action<Mv
...
+            if (useEndpointRouting)
+            {
+                // @robertmclaws: This is SUPER expensive, so don't do it unless we need it.
+                //      https://github.com/dotnet/aspnetcore/blob/release/8.0/src/Http/Routing/src/DependencyInjection/RoutingServiceCollectionExtensions.cs
+                services.AddRouting();
+            }
 
-            //RWM: Make sure that Restier works in any situation without needing additional knowledge.
-            return services.AddRouting();
+            // @robertmclaws: Make sure that Restier works in any situation without needing additional knowledge.
```
robertmclaws, line 130: Don't forget to make these documentation changes in the other overloads.

PR #755, opened 2024-01-25
Observe the route prefix in RestierSwaggerProvider

### LINKED ISSUE #749: Restier Swagger not applying route prefix
i configured by Restier with a route prefix of 'odata' and this works when i use it directly but the swagger  ui does not seem to have that and errors.
example from swagger ui of wrong example:
`curl -X 'GET' \
  'https://localhost:7221/Addresses?%24top=100' \
  -H 'accept: application/json'
  ` 

it should be:
`curl -X 'GET' \
  'https://localhost:7221/odata/Addresses?%24top=100' \
  -H 'accept: application/json'`

also actual output shows :

` {
"@odata.context": "https://localhost:7221/odata/$metadata#Addresses",
"value": [
{
"Id": 1,
...
}
]
}`

### Assemblies affected
restier swagger 1.1.0

### Reproduce steps
use the sample for noerthwind and just add routprefix and try it.
from my code:
`app.UseEndpoints(endpoints =>
{
    endpoints.Select().Expand().Filter().OrderBy().MaxTop(100).Count().SetTimeZoneInfo(TimeZoneInfo.Utc);
    endpoints.MapRestier(builder =>
    {
        builder.MapApiRoute<RiaODataController>("v1", "odata", true);
    });
});`

### Expected result
swagger ui would show the route as 
` curl -X 'GET' \
  'https://localhost:7221/odata/Addresses?%24top=100' \
  -H 'accept: application/json' `
and the ui "try it" button would return data

### Actual result
http error 404 - not found

0xced:
## Before this commit

When mapping a Restier route with a non empty route prefix, the generated swagger file does not include the route prefix in the `servers[0].url` property.

```csharp
app.MapRestier(restier => restier.MapApiRoute<EntityFrameworkApi<MyDbContext>>(routeName: "RestierDefault", routePrefix: "api"));
```

❌ The generated url is `http://localhost:5000`, making all requests from SwaggerUI fail with 404 errors.

## After this commit

The route prefix is included in the generated swagger file in the `servers[0].url` property.

✅ The generated url is `http://localhost:5000/api/`, making SwaggerUI usable.

Fixes #749

### DIFF
--- a/src/Microsoft.Restier.AspNetCore.Swagger/RestierSwaggerProvider.cs
+++ b/src/Microsoft.Restier.AspNetCore.Swagger/RestierSwaggerProvider.cs
@@ -56,11 +56,10 @@ namespace Microsoft.Restier.AspNetCore.Swagger
             openApiSettings?.Invoke(settings);
 
             // @robertmclaws: The host defaults internally to localhost; isn't set automatically.
-            var requestHost = httpContextAccessor.HttpContext.Request.Host.Value;
-            if (!settings.ServiceRoot.ToString().StartsWith(requestHost))
-            {
-                settings.ServiceRoot = new Uri($"{httpContextAccessor.HttpContext.Request.Scheme}://{requestHost}");
-            }
+            var request = httpContextAccessor.HttpContext?.Request ?? throw new InvalidOperationException("The HttpContext is not available");
+            var routePrefix = perRouteContainer.GetRoutePrefix(documentName);
+            var path = string.IsNullOrEmpty(routePrefix) ? "" : $"{routePrefix}/";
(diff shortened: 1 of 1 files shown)


robertmclaws:
Great catch, thanks!

robertmclaws:
@0xced FYI, if you're looking for a cleaner / more performant way to accomplish the same task in the future, I checked in an improvement here: https://github.com/OData/RESTier/blob/main/src/Microsoft.Restier.AspNetCore.Swagger/RestierSwaggerProvider.cs#L64-L71

HTH!

robertmclaws:
New package is live here: https://www.nuget.org/packages/Microsoft.Restier.AspNetCore.Swagger/1.1.1-CI-20240706-232326

LMK if you run into issues, otherwise will roll an RC build next week.

Issue #760, opened 2024-07-04
Batchhandler results in an error when doing bulk updates.


MitchKuijpers:
*Short summary (3-5 sentences) describing the issue.*  

### Assemblies affected
AspNet RESTier 1.1.0  

### Reproduce steps
Any batch request with more then 1 change.

### Expected result
Batch request is properly handled.

### Actual result
Batch crashes sometimes, depending on how many changes are sent. 

### Additional details
RestierBatchChangeSetRequestItem defines a dictionary that is used while firing the individual requests async.

PR #761, opened 2024-07-04
Batchhandler results in an error when doing bulk updates.

MitchKuijpers:
### Issues
*This pull request fixes issue #760 

### Description
Concurrent dictionary to resolve the crashes.

- [ ] *Test cases added*

### Additional work necessary
Hopefully It won't have more concurrency errors. The default method doesn't resolve the requests all at the same time, and I don't believe it is tested properly after this change.


MitchKuijpers:
@microsoft-github-policy-service agree company="Nocore"

robertmclaws:
Great catch, thanks!

robertmclaws:
Fixed here: https://www.nuget.org/packages/Microsoft.Restier.AspNetCore/1.1.1-CI-20240706-232326

LMK if there are any issues, otherwise an RC build will go out next week.

Issue #762, opened 2024-07-18
Batch Request does not take dependsOn into account


MartinTopfstedtSinc:
A batch request with 2 request, where the second request depends on the first. The first requests creates a new item and the second should create a new related item to the item from the first request, therefore the "$ContentId" url syntax is used.in the second request. Both requests are in the same atomicityGroup. 

The Problem is that both request get executed in parallel, and the second request can' t replace the "$ContentId" in the url because the first request didn't finish yet.

### Assemblies affected
Microsoft.Restier.AspNetCore 1.1.1

### Reproduce steps
A batch request contains 2 request, where the second request depends on the first. The first requests creates a new item and the second should create a new related item to the item from the first request, therefore the "$1" url syntax is used. Both requests are in the same atomicityGroup. 
*The simplest set of steps to reproduce the issue. If possible, reference a commit that demonstrates the issue.*  

### Expected result
The BatchChangeSetRequestItem should consider the dependsOn property of a request and order them correctly and execute only possible requests in parallel.

### Actual result
Both requests are executed in parallel.

### Additional details

For a small example see: https://docs.oasis-open.org/odata/odata-json-format/v4.01/odata-json-format-v4.01.html#sec_ReferencingNewEntities

https://github.com/OData/RESTier/blob/013bc0699cbf440dac78c9c07df30e8377d8b631/src/Microsoft.Restier.AspNet/Batch/RestierBatchChangeSetRequestItem.cs#L98

MitchKuijpers:
I think sending a singlechangset should never be async?
However for splitting into operations it should be fine...

jspuij:
This is an interesting issue, because I think it highlights some delicate formulations in the OData spec. (I used the formal OData 4.0 spec to cross reference).

Firstly, requests within a batch have to be processed in order. Requests within a change-set **MAY** be executed in any order (but it's not mandatory).

https://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part1-protocol.html#sec_ProcessingaMultipartBatchRequest

However, within a change-set previously inserted entities CAN be referenced by a content-id:

https://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part1-protocol.html#sec_ReferencingNewEntities

This of course implies, as @MartinTopfstedtSinc noted, that any request on a relational database that has foreign key constraints enabled cannot be executed in parallel.

@robertmclaws I think this is something we have to fix. Now there are multiple ways to go about this:

[...]

PR #767, opened 2024-10-29
Handle request in order.

MitchKuijpers:
### Issues
*This pull request fixes issue #762

### Description
I tried to find a method to check for the depends on id but could not find one. So the only fix for now would be resolving them in order. This is many times slower, but it's better then it being broken.

- [ ] *Test cases added*

### DIFF
--- a/src/Microsoft.Restier.AspNet/Batch/RestierBatchChangeSetRequestItem.cs
+++ b/src/Microsoft.Restier.AspNet/Batch/RestierBatchChangeSetRequestItem.cs
@@ -95,7 +95,10 @@ namespace Microsoft.Restier.AspNet.Batch
             // - the ChangeSet is submitted
             // - the responses are created and
             // - the controller actions have returned
-            await Task.WhenAll(responseTasks).ConfigureAwait(false);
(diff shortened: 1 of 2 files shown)


MartinTopfstedtSinc:
Does this work? I tried this myself but for some reason it waited inside the Controller for the other requests to finish until it populates the Response to the ChangeSet. (https://github.com/OData/RESTier/blob/013bc0699cbf440dac78c9c07df30e8377d8b631/src/Microsoft.Restier.AspNetCore/RestierController.cs#L194)

MitchKuijpers:
I'm not sure what you mean by that. I built it with this change and tested it both locally and in our test environment. However, it's really slow now. For that batches, but it works good for a single changeset where I don't want any entities saved when one fails.

That said, even if we want a "good" fix, we should only process the entities that have dependencies in the correct order. But I don't really see the need for this added complexity.

Personally, I would prefer providing a header to make the request asynchronous or perhaps creating two separate endpoints. I'm not entirely confident about what OData expects here. (I believe the standard Microsoft batch handler doesn't process requests in parallel, but what does it do with the dependsOn ID? Nothing?)

PR #774, opened 2026-04-15
Add YAML-based pipelines

gathogojr:
### Description
Adds YAML-based pipelines and uses latest code-signing infra


gathogojr:
Replaced by https://github.com/OData/RESTier/pull/775

